PR #9, opened 2015-12-26
Support reading 'image' from external YAML file

### LINKED ISSUE #7: Get 'image' from external YAML file
In some projects, there may already exist a canonical source for *the image* to use to build (e.g. [`.gitlab-ci.yml`](http://doc.gitlab.com/ce/ci/yaml/README.html)).

To avoid the need to duplicate this image name into `.scuba.yml` as well, allow scuba to reference another YAML file which specifies `image`.  It might look something like this:

**.gitlab-ci.yml**
```yaml
image: jdoe/builder:1.0
# ...
```

**.scuba.yml**
```yaml
image:
  source:                    # 'image' is a dictionary now, not a string
    path: .gitlab-ci.yml     # Load this external YAML file
    node: image              # And get the value of this node
```

Note that this solution has the following limitations which could be removed:
- Only external YAML files are supported
   - *What other types of files would contain a docker image name?*
- Only top-level keys in the external YAML file can be referenced
   - *Are there e.g. CI yaml file specifications that put the image name somewhere else?*
   - *Is there a mechanism, something like XPATH, that allows for "navigation" of a YAML file?*

Are there any other examples of cross-YAML document references?

Related Stack Overflow questions:
- [How can I include an YAML file inside another?](http://stackoverflow.com/questions/528281/how-can-i-include-an-yaml-file-inside-another)
- [Can YAML anchors/references occur across files/documents?](http://stackoverflow.com/questions/10717087/can-yaml-anchors-references-occur-across-files-documents)

/cc @combsjy

JonathonReinhart:
This closes #7.


JonathonReinhart:
Going to re-do this on a rebased branch.

PR #18, opened 2016-01-06
Don't use .scubainit when DOCKER_HOST is set

### LINKED ISSUE #17: .scubainit breaks everything on OSX
Trying the latest (742eccf5) on OSX with boot2docker doesn't work, at all. It just quietly exits with status 0:
```
$ scuba
$ echo $?
0
```

Commenting out the tempfile removal, and running the command that scuba would run, docker does the same thing, so Scuba is at least not "hiding" the error.

JonathonReinhart:
This better separates the remote vs. local execution logic.
When running a remote docker container (DOCKER_HOST is set), we don't do
any of the .scubainit stuff added in #13.

This fixes #17.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,13 @@ This project adheres to [Semantic Versioning](http://semver.org/).
 
 ## [Unreleased]
 ### Added
-- Set umask in container to the same as the host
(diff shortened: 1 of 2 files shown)


JonathonReinhart:
@jstoneburner confirms that this works correctly on OSX.

PR #52, opened 2016-05-11
scubainit: Don't add scubauser if env vars not set

### LINKED ISSUE #51: scubainit adds bogus entries with --root
When `--root` is passed to scuba, scubainit adds `-1` entries to the passwd files:
```
# grep scubauser /etc/passwd /etc/group /etc/shadow
/etc/passwd:scubauser:x:-1:-1:Scuba User:/:/bin/sh
/etc/group:scubauser:x:-1:
/etc/shadow:scubauser:x:::::::
```

JonathonReinhart:
Refactored out use_scuba_user() which includes the assertions.

This fixes #51

### DIFF
--- a/scubainit/scubainit.c
+++ b/scubainit/scubainit.c
@@ -42,6 +42,21 @@ static int m_uid = -1;
 static int m_gid = -1;
 static int m_umask = -1;
 
+
+/* Returns true if the scuba user should be used */
+static bool
+use_scuba_user(void)
+{
+    if (m_uid >= 0) {
+        assert(m_gid >= 0);
+        return true;
+    }
+
+    assert(m_gid == -1);
+    return false;
+}
+
+
 /**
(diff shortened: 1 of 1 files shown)


JonathonReinhart:
The setting of `umask` should not be inside the `use_scuba_user()` conditional block.

scubainit/scubainit.c
```diff
@@ -425,13 +431,30 @@ main(int argc, char **argv)
...
+    }
+
+    /* FUTURE: Call pre-su hook */
+
+    /* Handle the scuba user */
+    if (use_scuba_user()) {
+        if (change_user() < 0)
+            exit(99);
+
+        if (m_umask >= 0) {
```
JonathonReinhart: This block should be outside the `if (use_scuba_user())` conditional.

PR #67, opened 2016-07-13
Add support for enhanced aliases

### LINKED ISSUE #60: Enhanced aliases
Allow complex, multi-line aliases, using the [common script schema](https://github.com/JonathonReinhart/scuba/blob/master/doc/yaml-reference.md#common-script-schema), like this:

```yaml
image: foo
aliases:
  build:
    script:
      - echo "build script starting"
      - make -j4
      - other stuff
```

When an alias is used, a script will be generated from it and injected into the container for execution.

*If an alias is a multiline script, then no additional arguments will be allowed in the scuba invocation.*

A somewhat important change made by this is that *all aliases will now be executed by a shell*, instead of directly, as before. This alone may warrant a v2.0 bump, and we should then probably execute everything (even non-aliases) via the shell.

JonathonReinhart:
This closes #60

### DIFF
--- a/example/alias_multiline/.scuba.yml
+++ b/example/alias_multiline/.scuba.yml
@@ -1,0 +1,9 @@
+image: !from_yaml ../common.yml image
+aliases:
+  simple: echo 'a simple alias'
+
+  complex:
+    script:
+      - echo 'this is a complex script'
+      - date
+      - uname -a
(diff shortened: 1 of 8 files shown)


JonathonReinhart:
I want to add an example `.scuba.yml` before merging.

JonathonReinhart:
@combsjy How do you feel about these semantics, as detailed in #60? Particularly:

> *If an alias is a multiline script, then no additional arguments will be allowed in the scuba invocation.*

We talked about instead allowing the user to specify (in `.scuba.yml`) where the command-line arguments would be inserted, but this would possibly cause breaking behavior. No matter what alternative we might come up with, I think it's outside of the scope of this PR.

Issue #77, opened 2016-08-07
scubainit not included in wheels built on Travis


JonathonReinhart:
The weel distribution of [scuba 2.0.0a3](https://testpypi.python.org/pypi?:action=display&name=scuba&version=2.0.0a3) (a6f54abb135f62ef350ca666710537e1ec7b8c91) (built from [Travis Job #295.2](https://travis-ci.org/JonathonReinhart/scuba/jobs/150437287)) ([`scuba-2.0.0a3-py2.py3-none-any.whl`](https://testpypi.python.org/packages/81/96/0117441d47e0e000ee4f72ba5d7b59c12c6bce1dbe49787feeeace1e1802/scuba-2.0.0a3-py2.py3-none-any.whl#md5=692bcd31a7390fb5f403cdb7d01df28c)) is missing `scubainit`.

My [`.travis.yml`](https://github.com/JonathonReinhart/scuba/blob/a6f54abb135f62ef350ca666710537e1ec7b8c91/.travis.yml#L58) says:
```yaml
    distributions: "sdist bdist_wheel"
```

Their travis-ci/dpl tool's [pypi deploy provider](https://github.com/travis-ci/dpl/blob/5c3e22ec200e01cc723274a625ffc2ea0f2c03f8/lib/dpl/provider/pypi.rb#L96):
```ruby
      def push_app
        context.shell "python setup.py #{pypi_distributions}"
        context.shell "twine upload -r pypi dist/*"
        context.shell "rm -rf dist/*"
        context.shell "python setup.py upload_docs #{pypi_docs_dir_option} -r #{pypi_server}"
      end
```

JonathonReinhart:
### Test: `$ python setup.py sdist bdist_wheel`

```
$ python setup.py sdist bdist_wheel
running sdist
running egg_info
creating scuba.egg-info
writing requirements to scuba.egg-info/requires.txt
writing scuba.egg-info/PKG-INFO
writing top-level names to scuba.egg-info/top_level.txt
writing dependency_links to scuba.egg-info/dependency_links.txt
writing entry points to scuba.egg-info/entry_points.txt
writing manifest file 'scuba.egg-info/SOURCES.txt'
reading manifest file 'scuba.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
writing manifest file 'scuba.egg-info/SOURCES.txt'
warning: sdist: standard file not found: should have one of README, README.rst, README.txt

running check
[...]

PR #96, opened 2017-06-16
setup.py: Simplify and avoid unnecessary calls to make

JonathonReinhart:
In 422546f154, we worked around issue #77 where scubainit was not being
included in bdist_wheels. The actual fix however, was to simply set
include_package_data = True.

See https://github.com/pypa/setuptools/issues/1064

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,52 +1,29 @@
 #!/usr/bin/env python
 from __future__ import print_function
 import scuba.version
-from setuptools import setup
-from setuptools.command.build_py import build_py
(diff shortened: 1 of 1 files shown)


JonathonReinhart:
Merged right into master - CI doesn't help us here, because it only runs on `master`.

Confirmed working: https://travis-ci.org/JonathonReinhart/scuba/jobs/243511170#L3881:
```
copying build/lib/scuba/scubainit -> build/bdist.linux-x86_64/wheel/scuba
```

PR #99, opened 2017-10-09
Do not detect as a remote run if DOCKER_HOST is set to a unix:// socket

jacob733:
Certain use cases set DOCKER_HOST even if it points to a local docker instance. So a simple check if the variable is set is insufficient in some cases. Adding a check for if the variable is set to something using the unix:// protocol will make this filter more exact.

### DIFF
--- a/scuba/__main__.py
+++ b/scuba/__main__.py
@@ -148,7 +148,10 @@ class ScubaDive(object):
 
     @property
     def is_remote_docker(self):
-        return 'DOCKER_HOST' in os.environ
+        if 'DOCKER_HOST' in os.environ:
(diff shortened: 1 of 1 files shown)


JonathonReinhart:
This looks good to me. For my own curiosity, would you mind elaborating on the use cases which set `DOCKER_HOST` even on a local Docker instance?

jacob733:
For my case, it is for use with Jenkins, where an explicit DOCKER_HOST setting provides for a more readable configuration.

Note that if DOCKER_HOST is set to something like tcp://127.0.0.1... then this will still not be detected as a local docker. This is because such a setting might just as well be pointing to a local SSH port forward, which is one way to safely connect to a remote docker daemon.

PR #120, opened 2018-09-09
Add support for setting environment variable in .scuba.yml

### LINKED ISSUE #88: Using environment variables in .scuba.yml or passing them
Is it already possible – or do you consider it a valuable addition – to use environment variables in the `.scuba.yml` file?

For one, these could be already present in the shell where `scuba` is run. Additionally, I think some variables added by `scuba` might be helpful. (Is there a list of already used variables available somewhere, like the directory where `scuba` was started, path and directory to `.scubafile.yml` etc.)?

The goal would be do define an alias like `build: ./do-build.sh ${SCUBA_HOME} ${SOME_ENV}`.

With regard to passing environment variables into the container, what about a new configuration setting, either global or per-alias, that lists env vars to be passed?

```
image: ...
aliases:
    build: 
        command: ./do-build.sh ${TARGET}
        env:
            - THIS_VAR
            - THAT_VAR
        
env:
    - TARGET_ARCH
    - COMPILE_MODE
```

Note that we'd probably need to change the alias syntax somewhat or allow another structure to place the `env` key.

This would add `-e THIS_VAR -e THAT_VAR` to the `docker run` command when using the `build` alias, and add `-e TARGET_ARCH -e COMPILE_MODE` in any case.

Does that make sense?

JonathonReinhart:
Closes #88

### DIFF
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -139,7 +139,7 @@ def _process_environment(node, name):
     elif isinstance(node, dict):
         for k, v in node.items():
             if v is None:
-                v = os.getenv(k)
+                v = os.getenv(k, '')
             result[k] = str(v)
     elif isinstance(node, list):
         for e in node:
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -129,6 +129,30 @@ def _process_script_node(node, name):
     raise ConfigError("{0}: must be string or dict".format(name))
 
 
+def _process_environment(node, name):
+    # Environment can be either a list of strings ("KEY=VALUE") or a mapping
+    # Environment keys and values are always strings
+    result = {}
+
+    if not node:
+        pass
+    elif isinstance(node, dict):
+        for k, v in node.items():
+            if v is None:
+                v = os.getenv(k)
+            result[k] = str(v)
+    elif isinstance(node, list):
+        for e in node:
+            k, v = parse_env_var(e)
+            result[k] = v
+    else:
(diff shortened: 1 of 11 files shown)


JonathonReinhart:
@xanarin Would you care to review this?

scuba/config.py
```diff
@@ -129,25 +129,58 @@ def _process_script_node(node, name):
...
+    # Environment can be either a list of strings ("KEY=VALUE") or a mapping
+    # Environment keys and values are always strings
+    result = {}
+
+    if not node:
+        pass
+    elif isinstance(node, dict):
+        for k, v in node.items():
+            if v is None:
+                v = os.getenv(k)
```
xanarin: Currently if a variable is unable to be expanded by the host environment, the value in the container is set to the string `None`, not an empty string. I believe it's because `os.getenv` returns None [if the requested key doesn't exist](https://docs.python.org/3.7/library/os.html#os.getenv) and the value of `str(None) = "None"`.

xanarin: This behavior also exists in `utils.parse_env_vars`.

JonathonReinhart: Good catch. I must have been thinking that `os.getenv('NONEXIST')` would return an empty string if the key doesn't exist.

Luckily, the fix is easy:
```diff
-                v = os.getenv(k)
+                v = os.getenv(k, '')
```

JonathonReinhart: Fixed in ea9259b.

Issue #127, opened 2019-03-21
Concatenate lists imported from external YAML files


xanarin:
I have a use case where it would be extremely helpful to concatenate multiple lists from a different YAML file using Scuba's `!from_yaml` directive.

For example, I have the following Gitlab CI YAML configuration for my project:
```yaml
build_compA:
  stage: build
  tags:
    - docker
  script:
    - scons -C compA

build_compB:
  stage: build
  tags:
    - docker
  script: 
    - cd compB/
    - compB/gradlew -g .gradle -p compB lint
    - compB/gradlew -g .gradle -p compB test
```

Right now, I have two different Scuba aliases that reference these two CI jobs:
```yaml
aliases:
  build_compA: !from_yaml .gitlab-ci.yml build_compA
  build_compB: !from_yaml .gitlab-ci.yml build_compB
```

However, when I want to build both projects in the same bash invocation, I have to append multiple Scuba calls together:
```bash
$ scuba build_compA && scuba build_compB
```

---

It would make building all components simultaneously much easier if I could create an alias in my Scubafile that concatenated together both lists to create a single list to be run in the alias. This is a sample of what that could look like:
```yaml
aliases:
  build_all:
    script:
        - !from_yaml .gitlab-ci.yml build_compA
        - !from_yaml .gitlab-ci.yml build_compB
```
Or, maybe something more like:
```yaml
aliases:
  build_all:
    script: !from_yaml [.gitlab-ci.yml build_compA] [.gitlab-ci.yml build_compB]
```

Though I acknowledge that this could potentially lead to a user shooting themselves in the foot if they concatenate incompatible lists of commands, I think in some cases this functionality could be very helpful and make some Scubafiles and scuba invocations more succinct. Especially in cases where 7+ components need to be built at once, this would make the scuba invocation much simpler and eliminate the need to create a bash script just to build the project.

JonathonReinhart:
I really like your first suggestion:
```yaml
aliases:
  build_all:
    script:
        - !from_yaml .gitlab-ci.yml build_compA
        - !from_yaml .gitlab-ci.yml build_compB
```

This is good because it introduces no new concepts to the yaml schema. Each list element would itself be a list from the other yamal file.

Implementing this should be nothing more than flattening the nested list (something that SCons does frequently).

PR #128, opened 2019-03-22
Add support for nested scripts

JonathonReinhart:
This enables concatenation of scripts from multiple external yaml blocks, as suggested in #127.

### DIFF
--- a/doc/yaml-reference.md
+++ b/doc/yaml-reference.md
@@ -165,8 +165,92 @@ image: gcc:5.1
 image: !from_yaml .gitlab-ci.yml image
 ```
 
+Here's a more elaborate example which defines multiple aliases which correspond
+to jobs defined by `.gitlab-ci.yml`:
+
+**`.gitlab-ci.yml`**
+```yaml
+
+build_c:
+  image: gcc:5.1
+  script:
+    - make something
+    - make something-else
+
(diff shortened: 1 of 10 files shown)


xanarin:
Should the "Accessing external YAML content" section of `doc/yaml-reference.md` be updated to document this new functionality?

JonathonReinhart:
@xanarin What do you think about the updated docs?

doc/yaml-reference.md
```diff
@@ -165,8 +165,65 @@ image: gcc:5.1
...
+
+**`.scuba.yml`**
+```yaml
+image: not-specified (see issue #124)
+
+aliases
+  build_c:
+    image: !from_yaml .gitlab-ci.yml build_c.image
+    script: !from_yaml .gitlab-ci.yml build_c.script
+  build_py:
```
xanarin: I think it would be helpful here to show that (at least for Gitlab CI configuration files), one can directly import the entire job block:
```yaml
aliases:
  build_py: !from_yaml .gitlab-ci.yml build_py
```
For experienced users of `scuba` this pattern is obvious, but since this doc is targeted to newer users, it might be helpful to have an example of this pattern. I would either change this alias in the example, or add a third.

PR #137, opened 2019-09-04
config: Update `!from_yml` to handle complex keys

### LINKED ISSUE #135: `!from_yml` can't handle keys containing '.' characters
Due to how the "key" is parsed at [config.py](https://github.com/JonathonReinhart/scuba/blob/0244c81ec482d3c60202028bc075621447bc3ad1/scuba/config.py#L65) '.' characters must be escaped or somehow encapsulated.

TimothySprague:
Fixes #135

Specifically, this allows keys to contain escaped '.' characters.

It also updates `doc/yaml-reference.md` with a relevant example and adds
a unit test for the new functionality.

### DIFF
--- a/doc/yaml-reference.md
+++ b/doc/yaml-reference.md
@@ -250,6 +250,28 @@ aliases
       - !from_yaml .gitlab-ci.yml part2.script
 ```
 
+This example shows how to reference job names containing a `.` character.
+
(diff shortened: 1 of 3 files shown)


JonathonReinhart:
Thanks a lot, @TimothySprague!

PR #138, opened 2019-09-25
Add the ability to optionally specify a user/group instead of scubauser

kuhlmasp:
Allow the user to specify a user/group instead of `scubauser` by use of environment variables.


kuhlmasp:
@JonathonReinhart successfully tested locally for `-r`, default, and user & group specification (both individually, and together).

JonathonReinhart:
Thanks for the contribution, @kuhlmasp. I took this a bit further in #153 by having `scuba` always tell `scubainit` to create a user with the same name as the user invoking scuba.

PR #153, opened 2020-03-05
Use calling username/groupname instead of 'scubauser'

JonathonReinhart:
#70 stopped using `/scubaroot` and instead uses the same path in the container as outside.

This does the same with the current username (and groupname): Instead of using `scubauser`, Scuba will use the username and groupname of the user invoking scuba.

This builds on top of and supersedes #138.


JonathonReinhart:
I'm currently hesitating about the implementation of `make_homedir` because it might overlap with the invoking user's actual home directory...

Issue #147, opened 2020-01-29
Add Option to .scuba.yml to Run an Alias as Root


xanarin:
## Problem
Currently, one can run an alias as `root` in the container by specifying the `-r` or `--root` flag on the command line. However, there is not currently a way to specify this option in the `.scuba.yml` config file.

## Solution
A new option could be added to `.scuba.yml` to specify that a particular alias should run in the container as `root`:
```yaml
image: debian:buster
aliases:
  whoami:
    script:
      - whoami
  whoami_su:
    root: true
    script:
      - whoami
```

```bash
$ scuba whoami
scubauser

$ scuba whoami_su
root
```

EDIT: Remove ability to have top-level `root` node, as this could lead to unknowing misuse of Scuba and (AFAIK) doesn't have any current practical use-cases.

PR #152, opened 2020-02-25
Setup GitHub workflows

JonathonReinhart:
This MR drops Travis-CI in favor of the new [GitHub Actions](https://help.github.com/en/actions). The [current Travis configuration](https://github.com/JonathonReinhart/scuba/blob/f6511a20cc4267092ed5b8bf3179342c82ded77c/.travis.yml) is broken (#151). Rather than continue to deal with the sub-optimal Travis CI setup, give GitHub Actions a chance.

This establishes three workflows:
- [`build-test.yml`](https://github.com/JonathonReinhart/scuba/blob/e08e3da14526176ce2998a6c737b523f1cfb2fc7/.github/workflows/build-test.yml)
  - Build Python packages and run tests for all supported python versions
- [`publish-master.yml`](https://github.com/JonathonReinhart/scuba/blob/e08e3da14526176ce2998a6c737b523f1cfb2fc7/.github/workflows/publish-master.yml)
  - On pushes to `master`, publish to [Test PyPI](https://test.pypi.org/project/scuba/)
  - This uses the GitHub Run ID as a tag suffix
- [`publish-release.yml`](https://github.com/JonathonReinhart/scuba/blob/e08e3da14526176ce2998a6c737b523f1cfb2fc7/.github/workflows/publish-release.yml)
  - On creation of releases (which correspond to tags), publish to [PyPI](https://pypi.org/project/scuba/)

PR #159, opened 2020-03-25
Add ability to specify shell and "run as root" in .scuba.yml

### LINKED ISSUE #143: Scuba cannot function on Docker Images missing /bin/sh
[Kaniko](https://github.com/GoogleContainerTools/kaniko) is "a tool to build container images from a Dockerfile, inside a container or Kubernetes cluster." This makes it possible to build Docker images in Scuba and in Gitlab CI using a Docker runner.

However, the Docker image for Kaniko is **extremely** stripped down. The base image doesn't even have a shell in it, but the [debug image](https://github.com/GoogleContainerTools/kaniko#debug-image) does. However, the shell is in the `debug` image is located at `/busybox/sh`, and there is no `/bin/sh`. This causes Scuba to be non-functional because it requires `/bin/sh` to exec the `.scuba/command.sh` script:
```
$ scuba --version
scuba 2.3.0+49.g5bd095a

$ scuba --image gcr.io/kaniko-project/executor:debug echo "hi"
Error: unknown command "/bin/sh" for "executor"
Run 'executor --help' for usage.

$ sudo scuba --image gcr.io/kaniko-project/executor:debug --entrypoint="/busybox/sh" echo "hi"
/busybox/sh: can't open '/bin/sh': No such file or directory
```

To be able to use the Kaniko image with Scuba (and be able to build Docker images in Scuba), I think Scuba will need to have a flag (and maybe a `.scuba.yml` config option) to change which shell Scuba targets with the `command.sh` script.

xanarin:
Closes #143 and #147

### DIFF
--- a/scuba/__main__.py
+++ b/scuba/__main__.py
@@ -250,11 +250,15 @@ class ScubaDive(object):
         # with SELinux. See `man docker-run` for more information.
         self.vol_opts = ['z']
 
+        # Process any aliases
+        context = self.config.process_command(self.user_command,
+                image=self.image_override, shell=self.shell_override)
 
         # Pass variables to scubainit
         self.add_env('SCUBAINIT_UMASK', '{:04o}'.format(get_umask()))
 
-        if not self.as_root:
+        # Check if the CLI args specify "run as root", or if the command (alias) does
+        if not self.as_root and not context.as_root:
             uid = os.getuid()
             gid = os.getgid()
             self.add_env('SCUBAINIT_UID', uid)
@@ -271,18 +275,15 @@ class ScubaDive(object):
         # /usr, and Fedora 28 gives an AVC denial.
         scubainit_cpath = self.copy_scubadir_file('scubainit', self.scubainit_path)
 
+        # Hooks
+        for name in ('root', 'user', ):
+            self.__generate_hook_script(name, context.shell)
+
         # allocate TTY if scuba's output is going to a terminal
         # and stdin is not redirected
         if sys.stdout.isatty() and sys.stdin.isatty():
             self.add_option('--tty')
 
-        # Process any aliases
-        context = self.config.process_command(self.user_command,
-                image=self.image_override, shell=self.shell_override)
-
-        # Hooks
-        for name in ('root', 'user', ):
-            self.__generate_hook_script(name, context.shell)
 
         '''
         Normally, if the user provides no command to "docker run", the image's
--- a/scuba/__main__.py
+++ b/scuba/__main__.py
@@ -61,6 +61,7 @@ def parse_scuba_args(argv):
     ap.add_argument('--list-available-options', action=ListOptsAction,
             help=argparse.SUPPRESS)
     ap.add_argument('--image', help='Override Docker image')
+    ap.add_argument('--shell', help='Override shell used in Docker container')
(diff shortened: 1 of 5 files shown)


xanarin:
@JonathonReinhart I appear to be unable to assign this PR to you, but I'd appreciate it if you could take a look at it.

scuba/__main__.py
```diff
@@ -316,8 +319,8 @@ def __setup_native_run(self):
 
         # The user command is executed via a generated shell script
         with self.open_scubadir_file('command.sh', 'wt') as f:
-            self.docker_cmd += ['/bin/sh', f.container_path]
-            writeln(f, '#!/bin/sh')
+            self.docker_cmd += [context.shell, f.container_path]
+            writeln(f, '#!{}'.format(context.shell))
```
JonathonReinhart: We probably don't really need the she-bang line at all, since we're explicitly invoking the shell (via `self.docker_cmd`).

```suggestion
```

scuba/__main__.py
```diff
@@ -366,7 +369,7 @@ def __generate_hook_script(self, name):
 
             self.add_env('SCUBAINIT_HOOK_{}'.format(name.upper()), f.container_path)
 
-            writeln(f, '#!/bin/sh')
+            writeln(f, '#!{}'.format(shell))
```
JonathonReinhart: Just noting that for the hook scripts, we definitely *do* need the she-bang line since `scubainit` simply invokes them with `system`:

https://github.com/JonathonReinhart/scuba/blob/3bb7dcda5764ec9536e5c467bc93e43a8191e645/scubainit/scubainit.c#L429-L435

JonathonReinhart:
Dammit, the new GitHub Workflow-based CI I configured in #152 didn't run. The reason is the `build-test` workflow is configured to trigger only on `push` and not on `pull_request`:

https://github.com/JonathonReinhart/scuba/blob/3bb7dcda5764ec9536e5c467bc93e43a8191e645/.github/workflows/build-test.yml#L3

Let me see if I can re-configure that. Then I might have you rebase... stand by.

scuba/__main__.py
xanarin: Yeah, I agree. We'll have to re-work the tests, as they check the shebang line to validate that the shell was set correctly. I went that route because scuba's shell escaping prevents me from just running `scuba echo \$SHELL`, as `$SHELL` is escaped to `'$SHELL'` before being passed into the container. However, since the shebang isn't actually relevant here, it's probably best to change how we're validating the shell.

JonathonReinhart: I'm fine leaving it like this for now, unless you have an easy change in mind.

JonathonReinhart:
Okay, I merged #160. Can you rebase onto latest `master` and force-push your branch? Hopefully that should trigger CI to run.

xanarin:
I have rebased and force-pushed

scuba/__main__.py
xanarin: I added 73c9a69 to address this conversation. I think the new way of checking the shell is more correct/rigorous.

JonathonReinhart:
> I have rebased and force-pushed

Great. The checks ran this time and all lights are green. Thanks for the contribution!

PR #161, opened 2020-03-26
Implies -r if the user is root

TimothySprague:
Closes JonathonReinhart/scuba#94

### DIFF
--- a/scuba/__main__.py
+++ b/scuba/__main__.py
@@ -412,7 +412,7 @@ def run_scuba(scuba_args):
         scuba_args.command,
         docker_args = scuba_args.docker_args,
         env = scuba_args.env_vars,
-        as_root = scuba_args.root,
+        as_root = os.geteuid() == 0 or scuba_args.root,
         verbose = scuba_args.verbose,
         image_override = scuba_args.image,
         entrypoint = scuba_args.entrypoint,
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -314,6 +314,18 @@ class TestMain(TmpDirTestCase):
         assert_equal(gid, 0)
         assert_equal(groupname, 'root')
 
+    @mock.patch('os.geteuid', return_value=0)
+    def test_user_run_as_root(self, geteuid_mock):
+        '''Verify scuba run as root is identical to "scuba -r"'''
+
+        uid, username, gid, groupname = self._test_user()
+
+        assert_true(geteuid_mock.called)
+        assert_equal(uid, 0)
+        assert_equal(username, 'root')
+        assert_equal(gid, 0)
+        assert_equal(groupname, 'root')
+
     def test_user_root_alias(self):
         '''Verify that aliases can set whether the container is run as root'''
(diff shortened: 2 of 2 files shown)


JonathonReinhart:
Can you add a test? You'll probably want to mock `os.geteuid` to return `0` and run `whoami` or `id` in the container.

JonathonReinhart:
I didn't include any rationale in issue #94. Do you have a use-case for this?

And now that I see it, I'm kind of wondering why this is necessary, especially since scuba now runs as the invoking user name, ID, path, etc.

The current behavior (on master) is almost correct, except for these errors (any one of which would apply):
```
scubainit: Group "root" already exists in /etc/group
scubainit: User "root" already exists in /etc/passwd
scubainit: User "root" already exists in /etc/shadow
```

I'm not sure yet if it makes more or less sense, but with this patch to `scubainit` to avoid those harmless errors, running scuba as root works:
```diff
diff --git a/scubainit/scubainit.c b/scubainit/scubainit.c
index fa55103..8500773 100644
--- a/scubainit/scubainit.c
+++ b/scubainit/scubainit.c
@@ -113,6 +113,11 @@ add_group(const char *path, const char *name, unsigned int gid)
             break;
 
         if (strcmp(gr->gr_name, name) == 0) {
+            /* Identical group name+gid exists; odd, but no problem */
+            if (gr->gr_gid == gid) {
+                result = 0;
+                goto out;
+            }
             errmsg("Group \"%s\" already exists in %s\n", name, path);
             goto out;
         }
@@ -184,6 +189,11 @@ add_user(const char *path, const char *name, unsigned int uid, unsigned int gid,
             break;
 
         if (strcmp(pw->pw_name, name) == 0) {
+            /* Identical user name+uid exists; odd, but no problem */
+            if (pw->pw_uid == uid) {
+                result = 0;
+                goto out;
+            }
             errmsg("User \"%s\" already exists in %s\n", name, path);
             goto out;
         }
@@ -256,7 +266,8 @@ add_shadow(const char *path, const char *name)
             break;
 
         if (strcmp(sp->sp_namp, name) == 0) {
-            errmsg("User \"%s\" already exists in %s\n", name, path);
+            /* Already exists */
+            result = 0;
             goto out;
         }
     }
```

Aside from an implementation difference, there is another (potentially important!) difference:
- When passing `-r` or adding your change and running as `root`, scuba will not set `SCUBAINIT_UID` and friends, scubainit will think it doesn't have to change users, and therefore the **user hook will not run**.
- With my change, `SCUBAINIT_UID` will be set to `0`, scubainit will think it *is* changing users (see #163 for a rename that makes this more clear), and therefore the **user hook *will* run**.

It's a little bit strange, but I feel like the latter behavior is more correct. Thoughts?

TimothySprague:
I'd probably agree that the latter behavior is better.  I haven't personally had much use so far for the user hooks, but I wouldn't want to break any existing functionality.

JonathonReinhart:
I'm closing this in favor of #164.

Sorry this didn't get merged, but thank you for your contribution!

Issue #101, opened 2017-11-03
Bash completion needs updated to support running without .scuba.yml


combsjy:
With v2.1.0 it is possible to run `scuba` without a `.scuba.yml` file.  If you attempt to do this, tab completion will fail with an error:
```
.scuba.yml not found here or any parent up to mount point
```

JonathonReinhart:
How should scuba react in this situation? Should it only try to complete the `--image` argument?

TimothySprague:
This should be addressed by the use of `argcomplete` per #136.

JonathonReinhart:
Resolved by #162.

PR #162, opened 2020-03-30
Use argcomplete

### LINKED ISSUE #136: Use argcomplete
https://github.com/kislyuk/argcomplete

TimothySprague:
Closes #136 

There is something that still needs addressed though.  According to [the argcomplete README](https://github.com/kislyuk/argcomplete#global-completion), either global completion needs to be enabled and we need to add the string `PYTHON_ARGCOMPLETE_OK` to the top of the `__main__.py` *or* a user will need to add `eval "$(register-python-argcomplete scuba)"` to their `.bashrc` in order to activate `argcomplete` completion.

Do we want to handle this?  If so, how?


JonathonReinhart:
This looks neat.

I think we should probably remove the `--list-available-options` and ` --list-aliases` options from scuba too, since they were only there for tab completion.

JonathonReinhart:
I think we should add `PYTHON_ARGCOMPLETE_OK` since it hurts nothing and allows the user to decide how they want to enable completion (global or register per-module).

Then we need to present those two options (with a link to the argcomplete readme) in our readme.

scuba/__main__.py
```diff
@@ -69,9 +70,18 @@ def parse_scuba_args(argv):
     ap.add_argument('-v', '--version', action='version', version='scuba ' + __version__)
     ap.add_argument('-V', '--verbose', action='store_true',
             help="Be verbose")
+
+    def list_aliases_completer(**_):
+        try:
+            _, _, config = ScubaDive.get_paths_and_config()
+            return sorted(config.aliases)
+        except (ConfigNotFoundError, ConfigError):
+            return []
```
JonathonReinhart, line 79: I guess this probably fixes #101?

Would this be a usecase for [`argcomplete.warn`](https://kislyuk.github.io/argcomplete/#printing-warnings-in-completers)?

scuba/__main__.py
```diff
@@ -207,6 +217,12 @@ def __locate_scubainit(self):
         if not os.path.isfile(self.scubainit_path):
             raise ScubaError('scubainit not found at "{}"'.format(self.scubainit_path))
 
+    @staticmethod
+    def get_paths_and_config():
+        '''Helper function to find and load .scuba.yml
+        '''
+        top_path, top_rel = find_config()
+        return top_path, top_rel, load_config(os.path.join(top_path, SCUBA_YML))
```
JonathonReinhart, line 225: It seems like this `staticmethod` would be better as a normal function, located in the `config`  module. 

And to avoid the redundant `os.path.join`, `find_config` could be updated to return a third item, the path to `.scuba.yml` itself.

JonathonReinhart:
Another enhancement would be to add a completer to the `--image` option which lists all docker images.

scuba/__main__.py
TimothySprague: Yes (to fixes #101).

It it truly a warning to run without a `.scuba.yml`?  It seems like development is moving in the direction of that being a "normal" use case.

TimothySprague:
I considered adding a completer to `--image`.  I think it's a good idea.

I noticed we currently do a `Popen` or `call`.  Any particular reason not to use the `docker` package to directly access the API?

scuba/__main__.py
TimothySprague: By updating `find_config` that way, we wouldn't need *any* extra functions.  Sounds like a fine idea to me.

scuba/__main__.py
JonathonReinhart: Well, it's not warning-worthy to simply "run" scuba without `.scuba.yml`. But follow this logic:

- If you type `scuba <TAB>`, you're asking scuba to assist in completing your command line.
- Since you haven't provided anything for `args.command` yet, you must be looking for a "command" (and not args to that command)
- Since scuba can't know the available commands inside the image, the only thing it can offer are aliases.
- But it can't do that if there isn't a `.scuba.yml`.

So I think in the case where the user has provided nothing to `command` yet, then this is a warning.

And this conversation made me realize the we should really look at what's been provided to `command`. You currently have `def list_aliases_completer(**_):` but I'm assuming there is interesting context being passed to the completer.
- If you type `scuba anything <TAB>` (you've provided at least one thing to `args.command`), then scuba should not attempt to complete anything. In fact, we should *cancel* our completion (is that a thing?) and let bash go back to the normal behavior, suggesting files in the current directory.

JonathonReinhart:
> I noticed we currently do a `Popen` or `call`. Any particular reason not to use the `docker` package to directly access the API?

Simplicity, mostly.

- Scuba started off as a [simple wrapper script](https://github.com/JonathonReinhart/scuba/blob/v0.1.0/src/scuba) so there was no point in using the API.
- Scuba supports arbitrary docker command line arguments, which doesn't scale if scuba has to map those to API calls
- As for external dependencies, Scuba currently only requires `PyYAML` (important where additional dependencies means more packages to manually transfer/install)
- I didn't really want scuba to be anywhere in the loop of the Docker API version dependency loop -- I've seen churn on projects like `gitlab-runner` which uses the Docker API and keeps having to bump it to maintain compatibility

scuba/__main__.py
TimothySprague: That makes sense.

There is [context](https://kislyuk.github.io/argcomplete/#specifying-completers) with completers that we can use to do what you're suggesting.  To "cancel a completion", it looks like returning an empty list is sufficient.

TimothySprague:
Those are all sound reasons and it makes sense to me.

Just FYI though, these changes will add `argcomplete` as a dependency (as noted in the updated `requirements.txt` and `setup.py`)

JonathonReinhart:
Should we make `argcomplete` a soft dependency? In other words, `try/except ModuleNotFoundError`, and conditionally call `argcomplete.autocomplete(ap)`? I think the completer functions could still be attached unconditionally, and since they'll never be called without `argcomplete`, you can use `argcomplete` unconditionally inside of them.

TimothySprague:
I think that strikes a good balance if we're trying to keep the hard dependencies to a minimum.

We can use the `extras` features for `pip` and `requirements.txt` and then update the `README.md`.

scuba/__main__.py
TimothySprague: Addressed in TimothySprague/scuba@f48e2fd5

scuba/__main__.py
TimothySprague: Addressed in TimothySprague/scuba@f48e2fd5

JonathonReinhart:
I tried testing this on my Debian 10 machine, by installing [`python3-argcomplete (1.8.1-1)`](https://packages.debian.org/buster/python3-argcomplete) via `apt`, and it didn't work.

The problem was that the pip/wheel-generated console script wasn't recognized by the autocompleter as a script it should follow.

This was fixed in https://github.com/kislyuk/argcomplete/pull/241 which first appeared (https://github.com/kislyuk/argcomplete/commit/2559a349c846a65b2f1afe76ff9d048c9b45cd4d) in `v1.10.1`, which is newer than what Debian has.

So note: If you're installing Scuba with pip from a wheel (the recommended method), you need `argcomplete>=1.10.1`.

JonathonReinhart:
I pushed two commits that really simplify the way `get_images()` is implemented. I'm sorry you had to go through all of that effort... especially considering how much easier the Python API would make this.

requirements.txt
```diff
@@ -1,3 +1,4 @@
+argcomplete [ARGCOMPLETE]
```
JonathonReinhart, line 1: Where is this syntax documented? How do I use it? `requirements.txt` is only for setting up the development environment; `setup.py` is where the actual dependencies go.

README.md
```diff
@@ -25,6 +25,10 @@ To install:
 
     $ sudo pip install scuba
 
+To install with `argcomplete` (for [bash completion support](#bash-completion)):
+
+    $ sudo pip install scuba[ARGCOMPLETE]
```
JonathonReinhart, line 30: Shouldn't this be lowercase, to match the name in `extras_require` from `setup.py`?
```
    $ sudo pip install scuba[argcomplete]
```

JonathonReinhart: After testing both `pip install .[ARGCOMPLETE]` and `pip install .[argcomplete]`, it appears the "extras" are actually case-insensitive. However, I still think lowercase is correct.

README.md
```diff
@@ -80,6 +84,13 @@ Scuba defines the following environment variables in the container:
   where `.scuba.yml` was found.
 
 
+## Bash Completion
+Scuba supports command-line completion using the [`argcomplete` package](https://github.com/kislyuk/argcomplete).  Per
+the [`argcomplete` README](https://github.com/kislyuk/argcomplete#global-completion), command-line completion can be
+activated by running `eval "$(register-python-argcomplete scuba)"` manually *or* by using the
+`activate-global-python-argcomplete` script.
```
JonathonReinhart, line 91: It's not really scuba's responsibility to teach users how to use `argcomplete`, but I think we could help the user a little more here.

- [ ] Running `eval "$(register-python-argcomplete scuba)` will only enable completion for the current instance of the shell. What a user would *really* want to do is add that line to either `.bashrc`, or perhaps more appropriately, `~/.bash_completion`.
- [ ] Running `activate-global-python-argcomplete` as a normal user will fail, because it tries to write to `/etc/bash_completion.d/` by default.
- [ ] Running `activate-global-python-argcomplete --user` is probably better, but by itself is incomplete.
  - It will install the script as `~/.bash_completion.d/python-argcomplete`
  - `~/.bash_completion.d/*` are *not* sourced by default. However `~/.bash_completion` is, so the [solution](https://serverfault.com/a/831184) is to add this to `~/.bash_completion`:
    ```bash
    for bcfile in ~/.bash_completion.d/* ; do
      [ -f "$bcfile" ] && . $bcfile
    done
    ```

scuba/__main__.py
```diff
@@ -56,11 +62,11 @@ def parse_scuba_args(argv):
             help='Environment variables to pass to docker')
     ap.add_argument('--entrypoint',
             help='Override the default ENTRYPOINT of the image')
-    ap.add_argument('--list-aliases', action='store_true',
-            help=argparse.SUPPRESS)
-    ap.add_argument('--list-available-options', action=ListOptsAction,
-            help=argparse.SUPPRESS)
-    ap.add_argument('--image', help='Override Docker image')
+
+    def _list_images_completer(**_):
```
JonathonReinhart, line 66: Nit-pick: Can you move these two `_completer` functions up a little bit so we can keep all of the `ap.add_argument()` calls together? I get you were trying to keep them close to the argument that uses them, but I think it's more likely we'd want to look at all of the arguments.

scuba/__main__.py
```diff
@@ -69,9 +75,23 @@ def parse_scuba_args(argv):
     ap.add_argument('-v', '--version', action='version', version='scuba ' + __version__)
     ap.add_argument('-V', '--verbose', action='store_true',
             help="Be verbose")
+
+    def _list_aliases_completer(**kwargs):
+        # We don't want to try to complete any aliases if one was already given
+        if kwargs['parsed_args'].command:
+            return []
```
JonathonReinhart, line 82:
```suggestion
    def _list_aliases_completer(parsed_args, **kwargs):
        # We don't want to try to complete any aliases if one was already given
        if parsed_args.command:
            return []
```

scuba/config.py
```diff
@@ -73,20 +73,22 @@ def from_yaml(self, node):
 
 
 def find_config():
-    '''Search up the diretcory hierarchy for .scuba.yml
+    '''Search up the directory hierarchy for .scuba.yml
 
-    Returns: path, rel on success, or None if not found
+    Returns: path, rel, scuba_config on success, or None if not found
```
JonathonReinhart, line 78: This doesn't match the line added below.
```suggestion
    Returns: path, rel, config on success, or None if not found
```

scuba/config.py
```diff
@@ -73,20 +73,22 @@ def from_yaml(self, node):
...
     cross_fs = 'SCUBA_DISCOVERY_ACROSS_FILESYSTEM' in os.environ
     path = os.getcwd()
 
     rel = ''
     while True:
-        if os.path.exists(os.path.join(path, SCUBA_YML)):
-            return path, rel
+        config = os.path.join(path, SCUBA_YML)
+        if os.path.exists(config):
+            return path, rel, load_config(config)
```
JonathonReinhart, line 91: `config` might be a misleading name since we identified that as a "named" return value. How about:
```suggestion
        cfg_path = os.path.join(path, SCUBA_YML)
        if os.path.exists(cfg_path):
            return path, rel, load_config(cfg_path)
```

TimothySprague:
> This was fixed in [kislyuk/argcomplete#241](https://github.com/kislyuk/argcomplete/pull/241) which first appeared ([kislyuk/argcomplete@2559a34](https://github.com/kislyuk/argcomplete/commit/2559a349c846a65b2f1afe76ff9d048c9b45cd4d)) in `v1.10.1`, which is newer than what Debian has.
> 
> So note: If you're installing Scuba with pip from a wheel (the recommended method), you need `argcomplete>=1.10.1`.

Good catch, I typically do a `pip install` over `apt`, so I wouldn't have caught this.  I'll be sure to update `requirements.txt`.


> I pushed two commits that really simplify the way get_images() is implemented. I'm sorry you had to go through all of that effort... especially considering how much easier the Python API would make this.

Nice.  I'm not super familiar with all the `--format` values, so I tend to forget it exists (and end up doing things the hard way).

> Okay I think we're really close now. Do you want to fix these things up? If not, I can.

I'm fine to make the changes.

requirements.txt
TimothySprague: It's documents as [part of PEP 508](https://www.python.org/dev/peps/pep-0508/#extras), but I originally misread the usage.  I'll remove the `[ARGCOMPLETE]` tag.

README.md
TimothySprague: Good point.  I'll fix it.

scuba/config.py
TimothySprague: Good catch

README.md
TimothySprague: Okay.  I'll update this with some more explicit `argcomplete` activation examples.

scuba/config.py
TimothySprague: Addressed by TimothySprague/scuba@ea1be7f6fb2a25a586329f831373ff28e7a0a2db

requirements.txt
TimothySprague: Addressed by TimothySprague/scuba@ea1be7f6fb2a25a586329f831373ff28e7a0a2db

README.md
TimothySprague: Addressed by TimothySprague/scuba@ea1be7f6fb2a25a586329f831373ff28e7a0a2db

README.md
TimothySprague: Addressed by TimothySprague/scuba@ea1be7f6fb2a25a586329f831373ff28e7a0a2db

scuba/__main__.py
TimothySprague: Addressed by TimothySprague/scuba@ea1be7f6fb2a25a586329f831373ff28e7a0a2db

scuba/__main__.py
TimothySprague: Addressed by TimothySprague/scuba@ea1be7f6fb2a25a586329f831373ff28e7a0a2db

scuba/config.py
TimothySprague: Addressed by TimothySprague/scuba@ea1be7f6fb2a25a586329f831373ff28e7a0a2db

JonathonReinhart:
There's one last thing that kind of stinks: `argcomplete` seems to always suggest positional arguments, short options, and long options.

If I have a `.scuba.yml` file with `hello` and `goodbye` aliases, and type `scuba <TAB><TAB>`, I get these suggestions:
```
-d            --dry-run     --entrypoint  goodbye       hello         --image       -r            --shell       -V            --version     
--docker-arg  -e            --env         -h            --help        -n            --root        -v            --verbose
```
This isn't horrible, but it really hurts trying to find the alias you're looking for.

This is in contrast with the way the Bash completion works for `git`. With `git`, options are only suggested if you have first typed `-`.
```
$ git checkout <TAB><TAB>
(all branches and tags are suggested)
```
```
$ git checkout -<TAB>
(nothing suggested)
```
```
$ git checkout --<TAB>
--conflict=                   --ignore-skip-worktree-bits   --no-quiet                    --patch                       --recurse-submodules          
--detach                      --merge                       --orphan=                     --progress                    --theirs                      
--ignore-other-worktrees      --no-...                      --ours                        --quiet                       --track   
```
```
$ git checkout --no-<TAB>
--no-conflict                    --no-ignore-other-worktrees      --no-merge                       --no-patch                       --no-quiet                       --no-track 
--no-detach                      --no-ignore-skip-worktree-bits   --no-orphan                      --no-progress                    --no-recurse-submodules  
```
I wonder if there's a way to accomplish something similar with `argcomplete`...

PR #166, opened 2020-04-21
Run build/test workflow only on pull_request

JonathonReinhart:
Ideally we would run only on `push`, but that doesn't cover when someone creates/updates a PR on a fork, so we added the `pull_request` trigger in cf00c12f. That works well for forks.

However, that means when *I* push to a branch on my project with an associated PR, the tests all run twice.

Pushing to a branch without a PR, and expecting a build to run, seems unlikely enough that we can use `pull_request` exclusively for now.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -1,9 +1,7 @@
 name: Build and Test
 
-# Run this workflow whenever someone pushes to any branch,
-# or when a PR is created or pushed to.
+# Run this workflow whenever a PR is created or pushed to.
 on:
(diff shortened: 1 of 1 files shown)


JonathonReinhart:
WTF, I just created a PR for this branch, but the workflow did not run.

JonathonReinhart:
Okay maybe `[synchronize]` only runs when you *update* a PR? Ugh, why is this so difficult??

JonathonReinhart:
I need to close this and create a new PR for testing...

PR #177, opened 2021-07-17
Allow docker arguments in scubafile

### LINKED ISSUE #144: Allow Docker Arguments in Scubafile
Currently Scuba has the `-d` or `--docker-args` flag that allows the user to specify arguments that should be passed-through to docker. This is an essential feature for using Scuba in contexts where the container needs to forward ports from the host to the container (for instance, during web server development), or running in Docker's `privileged` mode to run applications that need KVM.

While the command-line argument is fantastic, the fact that there is no `.scuba.yml` counterpart means that instead of a developer just running `scuba run`, they have to do `scuba -d="-p 80:80 --privileged" run`. This also means that the additional arguments have to be documented and known by every developer, which increases friction for a developer on a new project. It would be much simpler if those arguments could be specified in the `.scuba.yml` file so that a new developer would be able to just run `scuba run` and have all of the additional arguments passed to Docker automatically.

The new `docker_args` key could either be specified at the top level, or inside of an alias:
```yaml
image: debian:buster
docker_args: --privileged

aliases:
  run_webserver:
    docker_args: -p 80:80 -p 443:443
    script:
        - python3 -m myKVMwebserver
```

Or, `docker_args` could take a list of arguments, more like:
```yaml
docker_args:
  - --privileged
  - -p 80:80
```
I dislike the list approach because it appears to be less readable to me.

I think this change would also (at least partially) address #140, #141, and #142.

adamjseitz:
Docker arguments can now be specified in the scubafile at the top-level or per-alias. The order of precendence for a single run is:

1. Arguments provided on command line via `-d`
2. Per-alias `docker_args` key
2. Top-level `docker_args` key

Explicitly providing a null or empty string value for `docker_args` at the alias level overrides the top-level `docker_args`, and results in no extra args passed to docker.

Resolves #144

### DIFF
--- a/doc/yaml-reference.md
+++ b/doc/yaml-reference.md
@@ -109,9 +109,12 @@ aliases:
       - echo $FOO $BAR
 ```
 
-Aliases can override the top-level `docker_args`:
+Aliases can extend the top-level `docker_args`. The following example will
+produce the docker arguments `--privileged -v /tmp/bar:/tmp/bar` when executing
+the `example` alias:
+
 ```yaml
-docker_args: -v /tmp/foo:/tmp/foo
+docker_args: --privileged
 aliases:
   example:
     docker_args: -v /tmp/bar:/tmp/bar
@@ -119,6 +122,34 @@ aliases:
       - ls -l /tmp/
 ```
 
+Aliases can also opt to override the top-level `docker_args`, replacing it with
+a new value. This is achieved with the `!override` tag:
+
+```yaml
+docker_args: -v /tmp/foo:/tmp/foo
+aliases:
+  example:
+    docker_args: !override -v /tmp/bar:/tmp/bar
+    script:
+      - ls -l /tmp/
+```
+
+The content of the `docker_args` key is re-parsed as YAML in order to allow
+combining the `!override` tag with other tags; however, this requires quoting
+the value, since YAML forbids a plain-style scalar from beginning with a `!`
+(see [the spec](https://yaml.org/spec/1.2/spec.html#id2788859)). In the next
+example, the top-level alias is replaced with an explicit `!!null` tag, so
+that no additional arguments are passed to docker when executing the `example`
+alias:
+
+```yaml
+docker_args: -v /tmp/foo:/tmp/foo
+aliases:
+  example:
+    docker_args: !override '!!null'
+    script:
+      - ls -l /tmp/
+```
 
 ### `hooks`
 
--- a/doc/yaml-reference.md
+++ b/doc/yaml-reference.md
@@ -40,6 +40,15 @@ environment:
   - SECRET
 ```
 
+### `docker_args`
+
+The optional `docker_args` node allows additional docker arguments to be
+specified.
+
+Example:
+```yaml
+docker_args: --privileged
+```
 
 ### `aliases`
 
@@ -88,6 +97,16 @@ aliases:
       BAR: "New"
     script:
       - echo $FOO $BAR
+```
+
+Aliases can override the top-level `docker_args`:
+```yaml
+docker_args: -v /tmp/foo:/tmp/foo
+aliases:
+  example:
+    docker_args: -v /tmp/bar:/tmp/bar
+    script:
+      - ls -l /tmp/
 ```
 
 
--- a/doc/yaml-reference.md
+++ b/doc/yaml-reference.md
@@ -47,7 +47,17 @@ specified.
 
 Example:
 ```yaml
-docker_args: --privileged
+docker_args: --privileged -v "/tmp/hello world:/tmp/hello world"
+```
+
+The value of `docker_args` is parsed as shell command line arguments using
+[`shlex.split`](https://docs.python.org/3/library/shlex.html#shlex.split).
+
+The previous example could be equivalently written in YAML's [single-quoted
+style](https://yaml.org/spec/1.2/spec.html#id2788097):
+
+```yaml
+docker_args: '--privileged -v "/tmp/hello world:/tmp/hello world"'
 ```
 
 ### `aliases`
--- a/scuba/__main__.py
+++ b/scuba/__main__.py
@@ -131,7 +131,8 @@ class ScubaDive:
         # These will be added to docker run cmdline
         self.env_vars = env
         self.volumes = []
-        self.options = docker_args or []
+        self.options = []
+        self.docker_args = docker_args or []
         self.workdir = None
 
         self.__locate_scubainit()
@@ -414,6 +415,12 @@ class ScubaDive:
             args += ['-w', self.workdir]
 
         args += self.options
+
+        # Precedence: (1) command-line -d, (2) alias docker_args, (3) top-level docker_args
+        if self.docker_args:
+            args += self.docker_args
+        elif self.context.docker_args is not None:
+            args += self.context.docker_args
 
         # Docker image
         args.append(self.context.image)
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -150,17 +150,15 @@ def _process_environment(node, name):
 
     return result
 
-def _get_entrypoint(data):
+def _get_nullable_str(data, key):
     # N.B. We can't use data.get() here, because that might return
-    # None, leading to ambiguity between entrypoint being absent or set
+    # None, leading to ambiguity between the key being absent or set
     # to a null value.
     #
     # "Note that a null is different from an empty string and that a
     # mapping entry with some key and a null value is valid and
     # different from not having that key in the mapping."
     #   - http://yaml.org/type/null.html
-    key = 'entrypoint'
-
     if not key in data:
         return None
 
@@ -175,9 +173,17 @@ def _get_entrypoint(data):
                 key, type(ep).__name__))
     return ep
 
+def _get_entrypoint(data):
+    return _get_nullable_str(data, 'entrypoint')
+
+def _get_docker_args(data):
+    args = _get_nullable_str(data, 'docker_args')
+    if args is not None:
+        args = shlex.split(args)
+    return args
 
 class ScubaAlias:
-    def __init__(self, name, script, image, entrypoint, environment, shell, as_root):
+    def __init__(self, name, script, image, entrypoint, environment, shell, as_root, docker_args):
         self.name = name
         self.script = script
         self.image = image
@@ -185,6 +191,7 @@ class ScubaAlias:
         self.environment = environment
         self.shell = shell
         self.as_root = as_root
+        self.docker_args = docker_args
 
     @classmethod
     def from_dict(cls, name, node):
@@ -194,9 +201,11 @@ class ScubaAlias:
         environment = None
         shell = None
         as_root = False
+        docker_args = None
 
         if isinstance(node, dict):  # Rich alias
             image = node.get('image')
+            docker_args = _get_docker_args(node)
             entrypoint = _get_entrypoint(node)
             environment = _process_environment(
                     node.get('environment'),
@@ -204,14 +213,14 @@ class ScubaAlias:
             shell = node.get('shell')
             as_root = node.get('root', as_root)
 
-        return cls(name, script, image, entrypoint, environment, shell, as_root)
+        return cls(name, script, image, entrypoint, environment, shell, as_root, docker_args)
 
 class ScubaContext:
     pass
 
 class ScubaConfig:
     def __init__(self, **data):
-        optional_nodes = ('image','aliases','hooks','entrypoint','environment','shell')
+        optional_nodes = ('image','aliases','hooks','entrypoint','environment','shell','docker_args')
 
         # Check for unrecognized nodes
         extra = [n for n in data if not n in optional_nodes]
@@ -222,6 +231,7 @@ class ScubaConfig:
         self._image = data.get('image')
         self._shell = data.get('shell', DEFAULT_SHELL)
         self._entrypoint = _get_entrypoint(data)
+        self._docker_args = _get_docker_args(data)
         self._load_aliases(data)
         self._load_hooks(data)
         self._environment = self._load_environment(data)
@@ -277,6 +287,10 @@ class ScubaConfig:
     def shell(self):
         return self._shell
 
+    @property
+    def docker_args(self):
+        return self._docker_args
+
 
     def process_command(self, command, image=None, shell=None):
         '''Processes a user command using aliases
@@ -297,6 +311,7 @@ class ScubaConfig:
         result.environment = self.environment.copy()
         result.shell = self.shell
         result.as_root = False
+        result.docker_args = self.docker_args
 
         if command:
             alias = self.aliases.get(command[0])
@@ -314,6 +329,8 @@ class ScubaConfig:
                     result.shell = alias.shell
                 if alias.as_root:
                     result.as_root = True
+                if alias.docker_args is not None:
+                    result.docker_args = alias.docker_args
 
                 # Merge/override the environment
                 if alias.environment:
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -1,3 +1,4 @@
+import collections
 import os
 import yaml
 import re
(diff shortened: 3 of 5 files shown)


scuba/__main__.py
```diff
@@ -415,6 +416,12 @@ def get_docker_cmdline(self):
 
         args += self.options
 
+        # Precedence: (1) command-line -d, (2) alias docker_args, (3) top-level docker_args
+        if self.docker_args:
+            args += self.docker_args
+        elif self.context.docker_args is not None:
+            args += self.context.docker_args
+
```
JonathonReinhart, line 424: Given that other options (e.g. `image`) respect this same precedence, this makes sense. But I could also see an additive behavior being useful; IOW the resulting set of docker args would be the union of the three.

There is some precedent for "merging" in the alias `environment` which is [combined with](https://github.com/JonathonReinhart/scuba/blob/v2.7.0/scuba/config.py#L329-L331) the top-level `environment`.

Starting with just `.scuba.yml`, I was hoping there would be some YAML syntax which would allow for explicitly pulling the top-level `docker_args` into the per-alias ones, but a) it's a string, not a list, and b) [there is no mechanism for lists](https://stackoverflow.com/questions/9254178) anyway.

**Rationale:**
- At least in the way I use scuba, things in `.scuba.yml` are "permanent" while things on the command line are one-offs for testing, etc.  I can imagine use cases where I want to add additional docker args (e.g. another bind-mount), but I can't think of any docker args that I would specifically want to omit/override.
- If scuba were additive but one needed to override for a one-off, I feel like it would be easier to comment-out the line in `.scuba.yml`, rather than the other way around

Thoughts?

tests/test_config.py
```diff
@@ -758,3 +809,118 @@ def test_alias_entrypoint(self):
...
+            f.write(r'''
+                image: na
+                docker_args: --privileged -v /tmp/:/tmp/
+                ''')
+
+        config = scuba.config.load_config('.scuba.yml')
+        assert config.docker_args == ['--privileged', '-v', '/tmp/:/tmp/']
+
+    def test_alias_docker_args_null(self):
+        '''docekr_args can be set to null via alias'''
```
JonathonReinhart, line 881:
```suggestion
        '''docker_args can be set to null via alias'''
```

tests/test_config.py
```diff
@@ -758,3 +809,118 @@ def test_alias_entrypoint(self):
...
+                    docker_args:
+                    script:
+                      - ugh
+                ''')
+
+        config = scuba.config.load_config('.scuba.yml')
+        assert config.aliases['testalias'].docker_args == []
+
+    def test_alias_docker_args_empty_string(self):
+        '''docekr_args can be set to empty string via alias'''
```
JonathonReinhart, line 897:
```suggestion
        '''docker_args can be set to empty string via alias'''
```

tests/test_config.py
```diff
@@ -758,3 +809,118 @@ def test_alias_entrypoint(self):
...
+                    docker_args: ''
+                    script:
+                      - ugh
+                ''')
+
+        config = scuba.config.load_config('.scuba.yml')
+        assert config.aliases['testalias'].docker_args == []
+
+    def test_alias_docker_args_set(self):
+        '''docekr_args can be set via alias'''
```
JonathonReinhart, line 913:
```suggestion
        '''docker_args can be set via alias'''
```

JonathonReinhart (commented):
Nice work, @adamjseitz. I left a few comments; let me know what you think.

Also, it seems like the ability to mount additional volumes is your primary motivator. I feel like whatever behavior is decided here would be equivalently applicable to a purpose-built `volumes` key.

scuba/__main__.py
adamjseitz: I was on the fence whether to go with the additive or override approach. I think I'm on board with reworking it as additive after reading though your thoughts.

adamjseitz: Thinking about this more, I think it would be nice to support both behaviors. 

Consider this (contrived) example where scuba is used to run tests for some software that is designed with "recommended" and "minimum" system requirements, and it takes advantage of the `--memory` docker argument to enforce such limits during test runs:
 
 ```yaml
image: debian:buster
# By default, we want to run with 2GB of memory - our software's "recommended" system requirements.
docker_args: --memory 2G
 
aliases:
  test-quick:
    script: ./run_quick_tests.sh

  test-all:
    script: ./run_all_tests.sh

  # Run tests in a container with limited memory to make sure we support low-end hardware
  test-mem-limited:
    docker_args: --memory 512M
    script: ./run_all_tests.sh
```

With the merging strategy, this use case of scuba would need to specify their default `--memory` argument on every alias , even if most aliases just wanted the same configuration. Trivial in this example, but easy to imagine this blowing up if there were many more aliases, and if the default args were significantly more complex (maybe they're also tweaking other hardware configs like CPU, read/write speeds, etc).

 A possible solution would be to support both a "simple" and "extended" `docker_args` definition. An alias could specify `docker_args` as a string, where it would be used directly, or as a dict with additional options:
 
 ```yaml
 image: debian:buster
 docker_args: --memory 2G
 
aliases:
  cmd1:
    # This alias ignores the global docker_args
    docker_args:
        strategy: override
        value: --memory 512M
  cmd2:
    # This alias uses the global docker_args
    docker_args:
        strategy: merge  # the default behavior - optional to specify here
        value: -p 80:80
  cmd3:
    # This alias also uses the global docker_args
    docker_args: -p 80:80 -p 443:443
 ```
 
With all the said, I think the interaction of args specified on the scuba CLI (`-d`) can only behave one way, and merging seems like the more reasonable default behavior.
 
 Let me know what you think on this.

Comment (JonathonReinhart) [reply scuba/__main__.py:424]: That's a great counterexample, and  a workable solution.

Here's another idea, inspired by something I saw elsewhere: We could use a `+` suffix on the key to denote an append behavior, and use replacement otherwise.

```yaml
docker_args: --memory 2G

aliases:
  test:
    docker_args: --memory 512M
  test2:
    docker_args+: --privileged
```

The two would be mutually exclusive.

It's a little less verbose than yours, and hopefully equivalently powerful.

Similarly, we could use e.g., `-d` and `-D` command line args.

Another random thought: we could support both shlex-split strings and lists, if we wanted to in the future.

xanarin: If I may drop my 2 cents here: I think having the ability to override the parent docker args is important for this feature, but I think it is the corner case. I expect that a majority of users are going to use/expect the appending behavior as opposed to the overriding behavior. To that end, perhaps it would make more sense to have the `docker_args` key (the one that is most logical and easy to remember) use the append behavior and use a key name of `docker_args-` or similar to use the override behavior.

adamjseitz: I don't think I am really a fan of inventing new syntax like a `+` or `-` character and strapping it on top of YAML, especially if it's only used for one specific key in all of scuba. 

The scuba file should prioritize readability, even if you're not familiar with scuba, over how fast you can type it. It is a configuration file, not a CLI command you have to type frequently.

scuba already has one special YAML construct: `!from_yaml`. Maybe sharing syntax with that would be best, and good middle ground between an overly verbose syntax and quick to write:

```yaml
image: debian:buster
docker_args: --memory 2G

aliases:
 cmd1:
   # This alias ignores the global docker_args
   docker_args: !override --memory 512M
 cmd2:
   # This alias uses the global docker_args
   docker_args:
       value: -p 80:80
```

It also plays nicely with allowing the default behavior to be append.

JonathonReinhart: Nice idea, @adamjseitz. I like it. I wonder if it will be possible to use both of them:

```
image: debian:buster
docker_args: --memory 2G

aliases:
  cmd1:
    # This alias ignores the global docker_args
    docker_args: !override !from_yaml some.yml in.here
```

adamjseitz: I took a first pass at implementing this. 

The implementation is a bit less clean than I would have liked, but it is complicated by the fact that `!override` is treated as a YAML tag, which is parsed first, and then scuba's key-specific `docker_args` parsing occurs later.

This mixin-class based solution allows the object produced by the YAML node to be still accessible with no special handling, but to allow detecting that an `!override` tag has been applied to that object. It does require some special handling around `None`, since it is not desirable to retype the `None` object.

> I wonder if it will be possible to use both of them:

It turns out it is, but with a caveat. The [YAML spec](https://yaml.org/spec/1.2/spec.html#id2788859) does not allow a the content of a plain-style scalar (regardless of the presense of a tag) to begin with a `!`. Thus, the combination is achieved with this syntax:

```yaml
image: debian:buster
docker_args: --memory 2G

aliases:
  cmd1:
    # This alias ignores the global docker_args
    docker_args: !override '!from_yaml some.yml in.here'
```

A potentially more useful application of this is to override with an explicit `!!null` value:

```yaml
image: debian:buster
docker_args: --memory 2G

aliases:
  cmd1:
    # This alias ignores the global docker_args
    docker_args: !override '!!null'
```

scuba/config.py
```diff
@@ -65,7 +91,31 @@ def from_yaml(self, node):
...
+            obj = OverrideNone()
+        else:
+            objtype = type(obj)
+            mixin_type = type('Override' + objtype.__name__, (objtype, OverrideMixin), dict())
+
+            try:
+                obj.__class__ = mixin_type
+            except TypeError:
+                # Primitive classes (e.g., int, str) don't support __class__ assignment
+                obj = mixin_type(obj)
```
JonathonReinhart, line 113: :clap:  Very cool technique!

I've seen recommendations *against* changing the type of an existing object (by assigning to `__class__`). However, I think most of the trouble can come from the fact that you may be missing functions/attributes that should be there if you actually instantiated the class. So in this case, I don't really see any problem with it, because the `OverrideMixin` doesn't have any methods and doesn't expect or *do* anything.

scuba/config.py
```diff
@@ -315,6 +387,11 @@ def process_command(self, command, image=None, shell=None):
...
                     result.as_root = True
 
+                if isinstance(alias.docker_args, OverrideMixin) or result.docker_args is None:
+                    result.docker_args = alias.docker_args
+                elif alias.docker_args is not None:
+                    result.docker_args.extend(alias.docker_args)
+
                 # Merge/override the environment
                 if alias.environment:
                     result.environment.update(alias.environment)
```
JonathonReinhart, line 397: Sorry for the scope creep, but should we handle `!override` here, also? It seems like a natural application for it, considering that the existing behavior is to extend. Maybe we handle this in a subsequent MR?

scuba/config.py
```diff
@@ -65,7 +91,31 @@ def from_yaml(self, node):
...
+            try:
+                obj.__class__ = mixin_type
+            except TypeError:
+                # Primitive classes (e.g., int, str) don't support __class__ assignment
+                obj = mixin_type(obj)
+
+        return obj
+
 Loader.add_constructor('!from_yaml', Loader.from_yaml)
+Loader.add_constructor('!override', Loader.override)
```
JonathonReinhart, line 118: Should we be concerned about, or provide protections against, users using `!override` in other, inappropriate contexts?

doc/yaml-reference.md
```diff
@@ -90,6 +109,47 @@ aliases:
...
+(see [the spec](https://yaml.org/spec/1.2/spec.html#id2788859)). In the next
+example, the top-level alias is replaced with an explicit `!!null` tag, so
+that no additional arguments are passed to docker when executing the `example`
+alias:
+
+```yaml
+docker_args: -v /tmp/foo:/tmp/foo
+aliases:
+  example:
+    docker_args: !override '!!null'
```
JonathonReinhart, line 149: You have so many tests, I hate to ask: Can we add a test for this?

JonathonReinhart (commented):
I think I really like how this `!override` looks. It's unfortunate there's so much going on behind it, but I think it's a clean and understandable implementation. And it has the added benefit that it should easily plug in to other places in the schema.

scuba/config.py
adamjseitz: I was thinking about this also. I think it makes the most sense to create an issue and bump it to future work.

scuba/config.py
adamjseitz: I don't think so. There are no protections against using the many tags built-in to PyYAML (see [here](https://pyyaml.org/wiki/PyYAMLDocumentation#yaml-tags-and-python-types)) in inappropriate contexts. Why should application-specific tags behave differently?

I think the only possible answer to this is that the `!override` will be silently ignored (and behave in ways the user doesn't expect), while using most of the other possible inappropriate tags would likely cause an error. However, It doesn't seem sustainable to make everything that accesses the YAML check for an `OverrideMixin` (and then raise an error), and building knowledge of which keys support `!override` into the `Loader` seems like misplaced logic.

doc/yaml-reference.md
adamjseitz: I had one at first, but I ended up removing it, since it's functionally the same as `test_alias_docker_args_override_from_yaml` - they both would be nesting another tag inside of an `!override` node. 

I'll put it in if you insist, but it seems like a "duplicate" to me.

JonathonReinhart: I agree that it's effectively the same test. My motivation for adding such a test are:
- I wasn't familiar with `!!null` so ensuring that it works as expected (and doesn't require e.g. opting into)
- Ensuring that the documented examples work as advertised

You've done more than enough, so don't feel obligated; I can address this later.

scuba/__main__.py
```diff
@@ -415,6 +416,12 @@ def get_docker_cmdline(self):
 
         args += self.options
 
+        # Precedence: (1) command-line -d, (2) alias docker_args, (3) top-level docker_args
+        if self.docker_args:
```
JonathonReinhart, line 420: In playing with this a little bit, I noticed that one cannot override `docker_args` from `.scuba.yml` to an empty list. This is because the argparse code uses `shlex.split()` on the argument string, and if given an empty string (`-d=''`), will produce an empty list, which makes this conditional False.

This is neither a showstopper, nor a problem with this PR, just something noteworthy.

JonathonReinhart:
There's also a bit of a security aspect to this addition. Consider a malicious code "publisher" trying to attack those who would download his open source project and build it.  (Granted, prior to Docker and scuba, one could easily drop something in a `Makefile` which runs on every developer's machine. However, this actually points out a benefit of using scuba: builds happen in a container so there's less risk).

Prior to this point, I'm not aware of anything that a publisher could put in their `.scuba.yml` file which could compromise the security of a consuming developer. Even if the image was untrusted, the docker model prevents anything from happening outside of the build directory.

Now with `docker_args`, one could add `--privileged` or `-v /:/hostroot` and a root hook can basically escalate privileges to host root, making things slightly *worse* than they were pre-scuba.

For trusted team environments this isn't as much of an issue, but for open-source software, this could be a problem. This concern is the motivation behind GDB's [`auto-load local-gdbinit`](https://sourceware.org/gdb/current/onlinedocs/gdb/Init-File-in-the-Current-Directory.html) and [`auto-load safe-path`](https://sourceware.org/gdb/current/onlinedocs/gdb/Auto_002dloading-safe-path.html) options.

To handle this problem in a similar way would require the addition of a user-local `scuba` configuration directory/file, with directives like "allow docker args" or "trusted paths". This is a lot of work for a fairly paranoid problem.

I'm fine merging this as-is, but wanted to take the opportunity to point out this impact.

scuba/__main__.py
```diff
@@ -131,7 +131,8 @@ def __init__(self, user_command, docker_args=None, env=None, as_root=False, verb
         # These will be added to docker run cmdline
         self.env_vars = env
         self.volumes = []
-        self.options = docker_args or []
+        self.options = []
+        self.docker_args = docker_args or []
```
JonathonReinhart, line 135: These aren't displayed in `__str__`.

doc/yaml-reference.md
JonathonReinhart: Actually, can you help me understand why using `!!null`:
```yaml
aliases:
  foo:
    docker_args: !override '!!null'
```

...is better than a simple empty string?
```yaml
aliases:
  foo:
    docker_args: !override
```

adamjseitz: Not necessarily better, but it's an instance of "explicit over implicit". It seems to me the `!!null` would be more obvious to a reader unfamiliar with the nuances in play as to what the intended result is.

scuba/__main__.py
JonathonReinhart: Added in 4768ee70513eebf49463e0bcf3e1a7dae3f79f43.

scuba/config.py
JonathonReinhart: Created issue #178.

PR #186, opened 2021-09-04
Add volumes to scubafile

### LINKED ISSUE #140: Mount additional volumes
Would it make sense to allow mounting additional directories through `.scuba.yml`, or is it already possible to do so?

My use case is that I would like to run tools to fetch dependencies (like `yarn`, `maven`, `composer`, ...). In order to access private repositories through SSH, one easy solution would be to bind-mount `$SSH_AUTH_SOCK` into the container.

Also, depending on the tool, it might be valuable to mount cache directories for `yarn` or `maven` into the build container.

Example:

```yaml
image: none

environment:
  SSH_AUTH_SOCK: /tmp/ssh-auth.sock

aliases:
  yarn:
    image: node:12-alpine
    mount:
      - $SSH_AUTH_SOCK:/tmp/ssh-auth.sock
      - $HOME/.cache/yarn:/root/.cache/yarn
```

Does that make sense and if so, could you guide me as to which changes are probably necessary? I am not a Python programmer. In particular, it would be helpful to know how I could run a locally checked-out version of `scuba` from inside my build directory.

JonathonReinhart:
This PR adds support for a new `volumes` key in `.scuba.yml` to bind-mount host directories into the container.

[**Latest docs build for `volumes` configuration**](https://scuba--186.org.readthedocs.build/en/186/configuration.html#volumes)

It supports both a simple form:
```yaml
volumes:
  /var/lib/foo: /host/foo
```

...and a complex form:
```yaml
volumes:
  /var/lib/foo:
    hostpath: /host/foo
    options: ro,cached
```

`volumes` can also be extended or overridden in an `alias`, as has now become the norm:
```yaml
volumes:
  /var/lib/foo: /host/foo
aliases:
  example:
    volumes:
      /var/lib/foo: /example/foo
      /var/lib/bar: /example/bar
    script:
      - ls -l /var/lib/foo /var/lib/bar
```

This closes #140.

### DIFF
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -6,6 +6,7 @@ import shlex
 
 from .constants import *
 from .utils import *
+from .dockerutil import make_vol_opt
 
 class ConfigError(Exception):
     pass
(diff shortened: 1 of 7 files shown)


scuba/scuba.py
```diff
@@ -318,6 +319,9 @@ def get_docker_cmdline(self):
         for hostpath, contpath, options in self.__get_vol_opts():
             args.append(make_vol_opt(hostpath, contpath, options))
 
+        for _, vol in self.context.volumes.items():
+            args.append(vol.get_vol_opt())
+
```
adamjseitz, line 324: Should the code here confirm that the configured volumes don't conflict with the volumes automatically created by scuba above?

adamjseitz (commented):
Overall looks good. This should allow much cleaner volume specifications with much more obvious override behavior than `docker_args` does.

scuba/config.py
```diff
@@ -239,9 +240,74 @@ def _get_docker_args(data):
...
+        #     hostpath: /host/foo
+        #     options: ro,z
+        if isinstance(node, dict):
+            hpath = node.get('hostpath')
+            if hpath is None:
+                raise ConfigError("Volume {} must have a 'hostpath' subkey".format(cpath))
+            return cls(
+                container_path = cpath,
+                host_path = hpath,
+                options = node.get('options').split(','),
```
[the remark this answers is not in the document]
JonathonReinhart: Great catch. Apparently I managed to always include `options` in all of my tests. Fixed.

scuba/scuba.py
JonathonReinhart: It's valid concern, but I think we can safely defer this.

The preconfigured volumes include `/.scuba` (for `scubainit` and the generated command script), and the absolute path of `.scuba.yml` (same inside/outside). I imagine it's fairly unlikely for someone to add a volume that matches one of those paths. If they do, and we pass conflicting `--volume` options, `docker run` will complain:
```
$ docker run --rm -it -v /tmp/a/:/var/lib/foo -v /tmp/b/:/var/lib/foo debian:10
docker: Error response from daemon: Duplicate mount point: /var/lib/foo.
See 'docker run --help'.
```

I would like to make the internal volumes use the same mechanics, but that's going to take a bit more work (due to the order that things are set up, particularly `self.vol_opts`).

Issue #191, opened 2022-01-10
Expand environmental variables in volume paths


xanarin:
A limitation of the solution that addressed #140 was that it did not support expanding environmental variables in volume parameters. This is important for use-cases where paths are different between each user's environment. For example, mapping in directories inside of a user's `/home` folder for user-specific configuration files. See [this comment](https://github.com/JonathonReinhart/scuba/issues/140#issuecomment-913812915) for more justification.

I would like Scuba to support a configuration like this:
```yaml
image: debian:buster
aliases:
    ex2:
        volumes:
            $HOME/.config/art: $HOME/.config/art
        script:
            - cat $HOME/.config/art/config.yml
```

xanarin:
This should have been closed when #192 was merged

PR #192, opened 2022-01-10
Expand env variables in volume paths in configuration file

xanarin:
For more context on this change, see #191.

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -166,6 +166,19 @@ of volume options:
         hostpath: /host/foo
         options: ro,cached
 
+The keys used in volume mappings can contain environment variables **that are
+expanded in the host environment**. For example, this configuration would map
+the user's ``/home/username/.config/application1`` directory into the container
+at the same path.
+
+.. code-block:: yaml
+
+    volumes:
+      $TEST_HOME/.config/application1: $TEST_HOME/.config/application1
+
+Note that because variable expansion is now applied to all volume keys, if one
+desires to have a key with an explicit ``$`` character, it must be written as
+``$$``.
 
 
 .. _conf_aliases:
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -268,9 +268,13 @@ def _get_volumes(data):
 def _expand_path(in_str):
     try:
         output = expand_env_vars(in_str)
-    except (KeyError, ValueError):
+    except KeyError as ke:
         # pylint: disable=raise-missing-from
-        raise ConfigError("Unset environmental variable used in '{}'".format(in_str))
+        raise ConfigError("Unset environment variable '{}' used in '{}'".format(ke.args[0], in_str))
+    except ValueError as ve:
+        raise ConfigError("Unable to expand string '{}' due to parsing "
+                "errors".format(in_str)) from ve
+
     return output
 
 class ScubaVolume:
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -261,6 +261,7 @@ def _get_volumes(data):
 
     vols = {}
     for cpath, v in voldata.items():
+        cpath = os.path.expandvars(cpath)
         vols[cpath] = ScubaVolume.from_dict(cpath, v)
     return vols
 
@@ -282,7 +283,7 @@ class ScubaVolume:
         if isinstance(node, str):
             return cls(
                 container_path = cpath,
-                host_path = node,
+                host_path = os.path.expandvars(node),
                 )
 
         # Complex form
@@ -296,7 +297,7 @@ class ScubaVolume:
                 raise ConfigError("Volume {} must have a 'hostpath' subkey".format(cpath))
             return cls(
                 container_path = cpath,
-                host_path = hpath,
+                host_path = os.path.expandvars(hpath),
                 options = _get_delimited_str_list(node, 'options', ','),
                 )
 
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -261,9 +261,17 @@ def _get_volumes(data):
 
     vols = {}
     for cpath, v in voldata.items():
-        cpath = os.path.expandvars(cpath)
+        cpath = _expand_path(cpath)
         vols[cpath] = ScubaVolume.from_dict(cpath, v)
     return vols
+
+def _expand_path(in_str):
+    try:
+        output = expand_env_vars(in_str)
+    except (KeyError, ValueError):
+        # pylint: disable=raise-missing-from
+        raise ConfigError("Unset environmental variable used in '{}'".format(in_str))
+    return output
 
 class ScubaVolume:
     def __init__(self, container_path, host_path=None, options=None):
@@ -283,7 +291,7 @@ class ScubaVolume:
         if isinstance(node, str):
             return cls(
                 container_path = cpath,
-                host_path = os.path.expandvars(node),
+                host_path = _expand_path(node),
                 )
 
         # Complex form
@@ -297,7 +305,7 @@ class ScubaVolume:
                 raise ConfigError("Volume {} must have a 'hostpath' subkey".format(cpath))
             return cls(
                 container_path = cpath,
-                host_path = os.path.expandvars(hpath),
+                host_path = _expand_path(hpath),
                 options = _get_delimited_str_list(node, 'options', ','),
                 )
 
--- a/scuba/utils.py
+++ b/scuba/utils.py
@@ -82,7 +82,7 @@ def writeln(f, line):
     f.write(line + '\n')
 
 def expand_env_vars(in_str):
-    """Expand environmental variables in a string
+    """Expand environment variables in a string
 
     Can raise `KeyError` if a variable is referenced but not defined, similar to
     bash's nounset (set -u) option"""
--- a/scuba/utils.py
+++ b/scuba/utils.py
@@ -1,6 +1,7 @@
 import errno
 import os
 from shlex import quote as shell_quote
+import string
 
 
 def shell_quote_cmd(cmdlist):
@@ -79,3 +80,10 @@ def get_umask():
 
 def writeln(f, line):
     f.write(line + '\n')
+
+def expand_env_vars(in_str):
+    """Expand environmental variables in a string
+
+    Can raise `KeyError` if a variable is referenced but not defined, similar to
+    bash's nounset (set -u) option"""
+    return string.Template(in_str).substitute(os.environ)
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -884,10 +884,10 @@ class TestConfig:
         assert v.host_path == '/host/bar'
         assert v.options == ['z', 'ro']
 
-    def test_volumes_with_env_vars_simple(self):
+    def test_volumes_with_env_vars_simple(self, monkeypatch):
         '''volume definitions can contain environment variables'''
-        os.environ["TEST_VOL_PATH"] = "/bar/baz"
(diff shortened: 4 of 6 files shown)


scuba/config.py
```diff
@@ -261,6 +261,7 @@ def _get_volumes(data):
 
     vols = {}
     for cpath, v in voldata.items():
+        cpath = os.path.expandvars(cpath)
```
JonathonReinhart, line 264: It's unfortunate that [`os.path.expandvars()`](https://docs.python.org/3/library/os.path.html#os.path.expandvars) is so lenient and doesn't provide a `strict` option:

> Malformed variable names and references to non-existing variables are left unchanged.

How do you feel about using this [alternative](https://stackoverflow.com/a/37800939/119527) based on [`string.Template`](https://docs.python.org/3/library/string.html#template-strings)?

```python
def expandvars(s):
    return string.Template(s).substitute(os.environ)
```

If we use this, we'll need to handle the `KeyError` that can be raised if a variable is not set. Perhaps we map it to a `ConfigError`?

I would like to defer this to a future enhancement, but it's far easier to make a program less strict in a future release than it is to make it more strict.

Either way, I like having the indirection of a single function where expansion (of any kind) is handled, rather than multiple calls to `os.path.expandvars()`. Perhaps `path_expand`?

JonathonReinhart (changes requested):

scuba/config.py
xanarin: Changed in 8a8c680. The code now uses `string.Template.substitute`, and provides both a general purpose expansion function in `utils.py` and a config-specific implementation (that raises the right exception) in `config.py`.

tests/test_config.py
```diff
@@ -883,3 +883,76 @@ def test_alias_volumes_set(self):
...
+        assert isinstance(v, scuba.config.ScubaVolume)
+        assert v.container_path == '/bar/baz/foo'
+        assert v.host_path == '/moo/doo/foo'
+        assert v.options == []
+
+    def test_volumes_with_env_vars_complex(self):
+        '''complex volume definitions can contain environmental variables'''
+        os.environ["TEST_HOME"] = "/home/testuser"
+        os.environ["TEST_TMP"] = "/tmp"
+        os.environ["TEST_MAIL"] = "/var/spool/mail/testuser"
```
JonathonReinhart, line 912: This actually modifies the process-global environment, which is not a good practice for unit tests, and can lead to confusing bugs.

The right way is to _mock_ the environment. Luckily we're using `pytest` which provides [`monkeypatch.setenv()`](https://docs.pytest.org/en/6.2.x/monkeypatch.html). See [this question](https://stackoverflow.com/q/38410878/119527).

This is already being used in `test_utils.py`:
https://github.com/JonathonReinhart/scuba/blob/04dc2c2c77de136bbe2caee8a1cbf2909564a610/tests/test_utils.py#L69

So after adding the fixture (by simply adding the parameter named `monkeypatch`):
```python
    def test_volumes_with_env_vars_complex(self, monkeypatch):
```
You can then:
```suggestion
        monkeypatch.setenv("TEST_HOME", "/home/testuser")
        monkeypatch.setenv("TEST_TMP", "/tmp")
        monkeypatch.setenv("TEST_MAIL", "/var/spool/mail/testuser")
```

tests/test_config.py
```diff
@@ -883,3 +883,76 @@ def test_alias_volumes_set(self):
         assert v.container_path == '/bar'
         assert v.host_path == '/host/bar'
         assert v.options == ['z', 'ro']
+
+    def test_volumes_with_env_vars_simple(self):
+        '''volume definitions can contain environmental variables'''
+        os.environ["TEST_VOL_PATH"] = "/bar/baz"
+        os.environ["TEST_VOL_PATH2"] = "/moo/doo"
```
JonathonReinhart, line 890:
```suggestion
        monkeypatch.setenv("TEST_VOL_PATH", "/bar/baz")
        monkeypatch.setenv("TEST_VOL_PATH2", "/moo/doo")
```

tests/test_config.py
```diff
@@ -883,3 +883,76 @@ def test_alias_volumes_set(self):
...
+        v = vols['/var/spool/mail/container']
+        assert isinstance(v, scuba.config.ScubaVolume)
+        assert v.container_path == '/var/spool/mail/container'
+        assert v.host_path == "/var/spool/mail/testuser"
+        assert v.options == ['z', 'ro']
+
+    def test_volumes_with_invalid_env_vars(self):
+        '''Volume definitions cannot include unset env vars'''
+        # Ensure that the entry does not exist in the environment
+        os.environ.pop("TEST_VAR1", None)
```
JonathonReinhart, line 951:
```suggestion
        monkeypatch.delenv("TEST_VAR1", raising=False)
```

JonathonReinhart (changes requested):
Thanks for the update. I found a few more issues.

scuba/config.py
```diff
@@ -261,9 +261,18 @@ def _get_volumes(data):
...
     vols = {}
     for cpath, v in voldata.items():
+        cpath = _expand_path(cpath)
         vols[cpath] = ScubaVolume.from_dict(cpath, v)
     return vols
 
+def _expand_path(in_str):
+    try:
+        output = expand_env_vars(in_str)
+    except (KeyError, ValueError):
```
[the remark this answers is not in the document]
xanarin: The [description](https://docs.python.org/3/library/string.html#string.Template.safe_substitute) for `string.Template.safe_substitute` includes the following:
>  Also, unlike with `substitute()`, any other appearances of the `$` will simply return `$` instead of raising `ValueError`.

The implementation of `string.Template.substitute` (taken here from the Python 3.10 source) also raises a ValueError exception in some cases:
```python3
def substitute(self, mapping=_sentinel_dict, /, **kws):
    if mapping is _sentinel_dict:
        mapping = kws
    elif kws:
        mapping = _ChainMap(kws, mapping)
    # Helper function for .sub()
    def convert(mo):
        # Check the most common path first.
        named = mo.group('named') or mo.group('braced')
        if named is not None:
            return str(mapping[named])
        if mo.group('escaped') is not None:
            return self.delimiter
        if mo.group('invalid') is not None:
            self._invalid(mo)
        raise ValueError('Unrecognized named group in pattern',
                         self.pattern)
    return self.pattern.sub(convert, self.template)
```

Therefore, I think we should catch both exceptions in `_expand_path`, though we may need to catch them individually to accommodate your request in https://github.com/JonathonReinhart/scuba/pull/192#discussion_r782752537 and provide good error messages for each one.

scuba/config.py
```diff
@@ -261,9 +261,18 @@ def _get_volumes(data):
...
+        # pylint: disable=raise-missing-from
+        raise ConfigError("Unset environmental variable used in '{}'".format(in_str))
```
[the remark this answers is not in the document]
xanarin: Huh...I never knew that!
more you know

Fixed in 83cccea

scuba/utils.py
```diff
@@ -79,3 +80,10 @@ def get_umask():
 
 def writeln(f, line):
     f.write(line + '\n')
+
+def expand_env_vars(in_str):
+    """Expand environmental variables in a string
```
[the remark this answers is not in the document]
xanarin: Fixed in 83cccea

tests/test_config.py
```diff
@@ -883,3 +883,76 @@ def test_alias_volumes_set(self):
         assert v.container_path == '/bar'
         assert v.host_path == '/host/bar'
         assert v.options == ['z', 'ro']
+
+    def test_volumes_with_env_vars_simple(self):
+        '''volume definitions can contain environmental variables'''
```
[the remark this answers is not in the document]
xanarin: Fixed in 83cccea

scuba/config.py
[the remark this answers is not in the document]
xanarin: Suggestion applied in 83cccea

tests/test_config.py
xanarin: In an earlier version of these tests I was modifying `$HOME`, which caused later tests to fail (but only because of the order the tests are executed in). I've been away from the world of high-level languages (with great tooling) too long, so I forgot about all of the great Python mocking features that exist to facilitate testing!

scuba/utils.py
```diff
@@ -79,3 +80,10 @@ def get_umask():
 
 def writeln(f, line):
     f.write(line + '\n')
+
+def expand_env_vars(in_str):
```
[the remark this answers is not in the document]
xanarin: Tests added in 2405ed8. Also, I added a test that raises a `ValueError` to document the condition for seeing that exception (and how to work around it if necessary).

tests/test_config.py
xanarin: Test now mock the environment in 2405ed8

tests/test_config.py
xanarin: Suggestion applied in 2405ed8

tests/test_config.py
xanarin: Suggestion applied in 2405ed8

xanarin:
I added documentation for this functionality change in 850f6b0.

The only backward-incompatability issue we may have is that if any users have the `$` character in their volume mapping strings that they **don't** want to be expanded, they will now have to write `$$`. However, given that the `volume` feature is pretty new and having `$` in the name of directories or files on Linux is a nightmare, I don't expect this to be an issue for anyone.

scuba/config.py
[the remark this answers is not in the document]
xanarin: As of 2405ed8 there is now a test that shows one condition where a `ValueError` can be raised from the `utils.expand_env_vars` function:
https://github.com/JonathonReinhart/scuba/blob/191-expand-vars-in-volumes/tests/test_utils.py#L147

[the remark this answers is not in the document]
JonathonReinhart: Nice catch! And crappy docs.

I confirmed that this code will raise `ValueError: Invalid placeholder in string: line 1, col 5`
```python
Template('foo $ bar').substitute()
```

scuba/utils.py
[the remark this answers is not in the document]
JonathonReinhart: Excellent!!

JonathonReinhart (approved):

JonathonReinhart:
Thanks a lot @xanarin for your patience and the high-quality code!

docs/configuration.rst
```diff
@@ -166,6 +166,19 @@ of volume options:
         hostpath: /host/foo
         options: ro,cached
 
+The keys used in volume mappings can contain environment variables **that are
```
JonathonReinhart, line 169: It's actually not just the keys (container path), but also the host path, which can contain environment vars (as shown in the example)

docs/configuration.rst
```diff
@@ -166,6 +166,19 @@ of volume options:
...
+expanded in the host environment**. For example, this configuration would map
+the user's ``/home/username/.config/application1`` directory into the container
+at the same path.
+
+.. code-block:: yaml
+
+    volumes:
+      $TEST_HOME/.config/application1: $TEST_HOME/.config/application1
+
+Note that because variable expansion is now applied to all volume keys, if one
```
JonathonReinhart, line 179: keys -> paths

JonathonReinhart: Opened #214 to fix

docs/configuration.rst
JonathonReinhart: Opened #214 to fix

PR #194, opened 2022-10-15
Add a command to initialize a default config

aedrax:
I've had a few people mention how it would be convenient if there were a built-in way to initialize a `.scuba.yml` for them to modify. I thought the change would be simple enough. Feel free to close if you feel it isn't necessary, no worries!

### DIFF
--- a/scuba/__main__.py
+++ b/scuba/__main__.py
@@ -49,6 +49,8 @@ def parse_scuba_args(argv):
     ap.add_argument('-d', '--docker-arg', dest='docker_args', action='append',
             type=lambda x: shlex.split(x), default=[],
             help="Pass additional arguments to 'docker run'")
+    ap.add_argument('-i', '--init', dest='init', action='store_true',
+            help="Initialize a new .scuba.yml in the current directory")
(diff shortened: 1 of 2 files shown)


JonathonReinhart:
Hey @aedrax, thanks for this and your other contributions to Scuba!

I've mulled over this a few times and I've decided to reject it. There's just no way there could ever be consensus on what the right fields are to include in this template.

- For the simplest use cases of Scuba, what this PR introduces is too complicated.
  - `shell` should rarely be used
  - `entrypoint` is rarely necessary
- It assumes the use of CMake, which alienates users of other build systems.
- It's debatably incorrect:
  - One can invoke `scuba` from any subdirectory of your project top-level directory (where `.scuba.yml`) lives. Yet, the `build` alias in this default creates a `build/` directory in the current dir. A safe implementation would use `$SCUBA_ROOT/build`.

We could attempt to fix these things, but I just don't see this being useful enough for the majority of users. Especially since configuring scuba isn't that complicated, in my opinion. IMHO this encourages the copy/paste/hack approach to configuration as opposed to deliberate decision-making.

I could be swayed, but there's insufficient justification for now.

PR #201, opened 2023-02-20
Create host paths for volumes before starting Docker

### LINKED ISSUE #195: volumes configuration defaults to root ownership if the file or directory at hostpath does not exist
Scuba's `volumes` configuration utilizes Docker's [bind-mount](https://docs.docker.com/storage/bind-mounts/) `--volume` feature which automatically creates `hostpath` as a directory if it doesn't exist when starting the container. This creates ownership problems because the Docker daemon creates the directory at `hostpath` as `root`.

**Reproduction configuration**
``` yaml
image: debian:11
volumes:
  ${PWD}/missing: ${PWD}/missing
aliases:
  test:
    script:
      - ls -ld missing
      - touch missing/test
```

``` bash
$ #rm -rf ./missing
$ scuba test
drwxr-xr-x 2 root root 4096 Dec  6 13:09 missing
touch: cannot touch 'missing/test': Permission denied
```

**Expected Behavior**
When starting a container with bind-mount volumes, Scuba should attempt to create a directory as the user when bind-mounting a non-existent `hostpath`.
1. If Scuba can create the directory without `root` it's **always** better, as both the current user and root will have access to the directory at `hostpath`.
2. If Scuba can't create the directory, then it's reasonable to fall back to Docker's default behavior without an error. Scuba should not modify file permissions at `hostpath`.

haboustak:
Docker creates host paths as root, which creates permission problems when volumes are within user-owned directories (like `$HOME`). If a host path doesn't exist, scuba duplicates Docker's behavior and attempts to create an empty directory at that location, but owned by the user that ran scuba.

Closes #195

### DIFF
--- a/scuba/scuba.py
+++ b/scuba/scuba.py
@@ -154,6 +154,30 @@ class ScubaDive:
             options = []
         self.volumes.append((hostpath, contpath, options))
 
+    def try_create_volumes(self):
+        '''Try to create non-existent host paths prior to docker run invocation
+
+        This only creates user-defined volumes from configuration. The scubadir
+        and the initial working directory either exist or are created as root by
+        Docker.
+        '''
+        # Cannot create local directories for a remote host
+        if self.is_remote_docker:
+            return
+
+        for vol in self.context.volumes.values():
+            if os.path.exists(vol.host_path):
+                continue
+
+            try:
+                # Create directories all the way to the target
+                os.makedirs(vol.host_path, exist_ok=True)
+            except PermissionError:
+                # Docker will create this path later as root
+                pass
+            except OSError as err:
+                raise ScubaError('Error creating volume host path: {}'.format(err)) from err
+
     def add_option(self, option):
         '''Add another option to the docker run invocation
         '''
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -949,3 +949,83 @@ class TestMain:
         out, _ = self.run_scuba(['doit'])
         out = out.splitlines()
         assert out == ["from the alias"]
+
+    def test_volumes_host_path_create(self):
+        '''Missing host paths should be created before starting Docker'''
+
+        userdir = Path('./user')
+        testfile = userdir / 'test.txt'
+
+        with open('.scuba.yml', 'w') as f:
+            f.write('''
+                image: {image}
+                volumes:
+                  /userdir: {user_dir}
+                '''.format(image = DOCKER_IMAGE,
+                           user_dir = userdir.absolute(),
+                           ))
+
+        self.run_scuba(['touch', '/userdir/test.txt'])
+
+        assert testfile.exists(), "Test file was not created"
+
+        info = userdir.stat()
+        assert info.st_uid != 0, "Directory is owned by root"
+        assert info.st_gid != 0, "Directory group is root"
+
+    def test_volumes_host_path_permissions(self):
+        '''Host path permission errors should be ignored'''
+
+        rootdir = Path('./root')
+        userdir = rootdir / 'user'
+        testfile = userdir / 'test.txt'
+
+        rootdir.mkdir()
+
+        with open('.scuba.yml', 'w') as f:
+            f.write('''
+                image: {image}
+                volumes:
+                  /userdir: {user_dir}
+                aliases:
+                   doit:
+                      root: true
+                      script: "touch /userdir/test.txt"
+                '''.format(image = DOCKER_IMAGE,
+                           user_dir = userdir.absolute(),
+                           ))
+
+        try:
+            # Prevent current user from creating directory
+            rootdir.chmod(0o555)
+            self.run_scuba(['doit'])
+        finally:
+            # Restore permissions to allow deletion
+            rootdir.chmod(0o755)
+
+        assert testfile.exists(), "Test file was not created"
+
+        info = userdir.stat()
+        assert info.st_uid == 0, "Directory is owned by root"
+        assert info.st_gid == 0, "Directory group is root"
+
+    def test_volumes_host_path_failure(self):
+        '''Host path failures due to OS errors prevent Docker run'''
+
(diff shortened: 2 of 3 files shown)


JonathonReinhart:
Hey Mike, upon a quick look, this seems fine. Can you add a test?

haboustak:
Yes, I can look into tests.

I also plan to review and update the documentation, because although it's a minor change in behavior, I think it's worth noting. I was struggling to effectively elucidate "docker volume host path created as the non-root user", so I left it for later.

scuba/scuba.py
```diff
@@ -154,6 +154,26 @@ def add_volume(self, hostpath, contpath, options=None):
             options = []
         self.volumes.append((hostpath, contpath, options))
 
+    def try_create_volumes(self):
+        '''Try to create non-existent host paths prior to docker run invocation
```
JonathonReinhart, line 158: It might be worthwhile to note that this only creates *user-configured* volumes (either via command-line option of via `.scuba.yml`) and not any scuba-internal volumes/mounts.

scuba/scuba.py
```diff
@@ -154,6 +154,26 @@ def add_volume(self, hostpath, contpath, options=None):
...
+            if os.path.exists(vol.host_path):
+                continue
+
+            try:
+                # Create directories all the way to the target
+                os.makedirs(vol.host_path, exist_ok=True)
+            except PermissionError:
+                # Docker will create this path later as root
+                pass
+            except Exception as err:
```
JonathonReinhart, line 174: Can we limit this to, say, `OSError`? Anything else is a likely programming mistake and a backtrace is warranted.

tests/test_main.py
```diff
@@ -949,3 +950,83 @@ def test_volumes_alias_override(self):
...
+                image: {image}
+                volumes:
+                  /userdir: {user_dir}
+                '''.format(image = DOCKER_IMAGE,
+                           user_dir = userdir.absolute(),
+                           ))
+
+        self.run_scuba(['touch', '/userdir/test.txt'])
+
+        assert os.path.exists(testfile.absolute()), "Test file was not created"
```
JonathonReinhart, line 971: Take advantage of the fact that `testfile` is a `Path`:
```suggestion
        assert testfile.exists(), "Test file was not created"
```

tests/test_main.py
```diff
@@ -949,3 +950,83 @@ def test_volumes_alias_override(self):
...
+
+        info = os.stat(userdir)
```
JonathonReinhart, line 973: Take advantage of the fat that `userdir` is a `Path`:
```suggestion
        info = userdir.stat()
```

tests/test_main.py
```diff
@@ -949,3 +950,83 @@ def test_volumes_alias_override(self):
...
+                   doit:
+                      root: true
+                      script: "touch /userdir/test.txt"
+                '''.format(image = DOCKER_IMAGE,
+                           user_dir = userdir.absolute(),
+                           ))
+
+        try:
+            # Prevent current user from creating directory
+            os.chmod(rootdir.absolute(), 0o555)
```
JonathonReinhart, line 1001:
```suggestion
            rootdir.chmod(0o555)
```

tests/test_main.py
```diff
@@ -949,3 +950,83 @@ def test_volumes_alias_override(self):
...
+            self.run_scuba(['doit'])
+        finally:
+            # Restore permissions to allow deletion
+            os.chmod(rootdir.absolute(), 0o755)
```
JonathonReinhart, line 1005:
```suggestion
            rootdir.chmod(0o755)
```

tests/test_main.py
```diff
@@ -949,3 +950,83 @@ def test_volumes_alias_override(self):
...
+
+        assert os.path.exists(testfile.absolute()), "Test file was not created"
```
JonathonReinhart, line 1007:
```suggestion
        assert testfile.exists(), "Test file was not created"
```

tests/test_main.py
```diff
@@ -949,3 +950,83 @@ def test_volumes_alias_override(self):
...
+
+        info = os.stat(userdir)
```
JonathonReinhart, line 1009:
```suggestion
        info = userdir.stat()
```

JonathonReinhart (changes requested):
Thanks for the PR, Mike. Overall looks good, just a few nits, mostly regarding `Path` methods.

I see this is still marked `Draft`. Is there anything else you're planning to change?

JonathonReinhart:
> Yes, I can look into tests.
> 
> I also plan to review and update the documentation, because although it's a minor change in behavior, I think it's worth noting. I was struggling to effectively elucidate "docker volume host path created as the non-root user", so I left it for later.

Looking at [the docs](https://scuba.readthedocs.io/en/latest/configuration.html#conf-volumes), it's never even mentioned that a non-existent volume path will be created. Scuba tries to be transparent to the user, and the whole point is to create files as the invoking user. Thus, I would be very surprised if anyone was relying on the current behavior (where Docker creates a nonexistent volume as `root`).

Given that, I wouldn't worry too much about the docs. If you want to add the fact that volumes are created (at all) as the invoking user, please feel free. And this will of course be a Fixed entry in the change log.

haboustak:
> I see this is still marked `Draft`. Is there anything else you're planning to change?

I left it as draft because I haven't tested this updated version with my actual use case: mounting a non-existent  `GOCACHE` directory into the container.  I'll address the feedback so far, run some final manual tests, and it should be ready for merge.

haboustak:
Feedback applied, rebased, and ready for merge. I've also tested real-world builds and am happy with the results.

JonathonReinhart (approved):

JonathonReinhart:
Note: The tests for Python 3.5 and 3.6 failed because `actions/setup-python@v1` seems to have bumped their minimum to Python 3.7. Time for #202.

PR #207, opened 2023-03-08
Add type annotations

### LINKED ISSUE #204: Implement typing and mypy support

JonathonReinhart:
This adds [PEP 484](https://peps.python.org/pep-0484/) type annotations to all files in the `scuba/` package, and enables type checking in CI via [mypy](https://mypy.readthedocs.io/).

Fixes #204.

### DIFF
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -1,12 +1,21 @@
-import collections
+from __future__ import annotations
+import dataclasses
 import os
-import yaml
 import re
 import shlex
+from typing import Any, List, Dict, Optional, TextIO, Tuple, Type, TypeVar, Union
+
+import yaml
+import yaml.nodes
 
 from .constants import *
 from .utils import *
 from .dockerutil import make_vol_opt
+
+CfgNode = Any
+CfgData = Dict[str, CfgNode]
+Environment = Dict[str, str]
+_T = TypeVar("_T")
 
 
 class ConfigError(Exception):
@@ -25,19 +34,17 @@ class OverrideMixin:
     and any object can be checked if it is an OverrideMixin using isinstance().
     """
 
-    pass
-
 
 class OverrideNone(OverrideMixin):
     """
     Represents a None value that also has Override behavior
     """
 
-    def __bool__(self):
+    def __bool__(self) -> bool:
         return False
 
 
-class OverrideList(collections.UserList, OverrideMixin):
+class OverrideList(list, OverrideMixin):
     pass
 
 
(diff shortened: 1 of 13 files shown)


scuba/config.py
```diff
@@ -103,15 +123,17 @@ def from_yaml(self, node):
...
         Implements !override constructor
         """
         # Load the content from the node, as a scalar
+        assert isinstance(node, yaml.nodes.ScalarNode)
         content = self.construct_scalar(node)
+        assert isinstance(content, str)
 
         # Dynamically add an OverrideMixin to the resulting object's type
-        obj = yaml.load(content, lambda s: Loader(s, root=self._root))
+        obj = yaml.load(content, self._rooted_loader(root=self._root))
```
JonathonReinhart, line 136: cc @adamjseitz: Just FYI since you implemented this in #177

adamjseitz: Thanks for the ping. Nice to see type annotations - and nice to see them doing their job, as it looks like it caught my previous mistake here.

JonathonReinhart: I'm not sure I'd call it a mistake as much as I would an overly-specific type annotation.

Typeshed [says](https://github.com/python/typeshed/blob/d74bea5e870708cb88a0ae8ca6f3e13ca5098531/stubs/PyYAML/yaml/__init__.pyi#L38) that the `Loader` arg is a `type[_Loader | _CLoader]`. But it is only [called](https://github.com/yaml/pyyaml/blob/957ae4d495cf8fcb5475c6c2f1bce801096b68a5/lib/yaml/__init__.py#L79) to create a loader instance. So I could argue that it should be relaxed and annotated as `Callable[[TextIO], _Loader | _CLoader]`, which your code would have satisfied.

Anyway, thanks for taking a look!

xanarin (approved):
I am not very familiar with Python type annotations, so I'm probably not the best person to check the correctness of the changes in this PR w.r.t. typing. However, all of the non-typing code changes look reasonable to me and since the tests are all still passing I'll conclude that there were no functionality regressions. There are still quite a few TODOs, but since they all concern typing I am comfortable with merging them into `main`.

Even though I'm not very experienced with the type annotations, I'm very glad to see that Scuba is moving to use them. I think this will make future modifications to Scuba by new developers easier (especially the code that parses and assembles config files) 👍🏻

JonathonReinhart:
> Even though I'm not very experienced with the type annotations, I'm very glad to see that Scuba is moving to use them.

I used to be slightly anti-typing since it felt like it went against the dynamic nature of Python. But then I worked in a codebase that required it, and learned to appreciate it.

> I think this will make future modifications to Scuba by new developers easier (especially the code that parses and assembles config files) 👍🏻 

Yep, exactly. #201 was the impetus for this. I had to chase a reference through 5 layers of function calls to figure out what its type was.

PR #227, opened 2023-09-13
Enable use of relative volume host-paths

### LINKED ISSUE #218: Detect or handle relative host paths in volumes
Currently a naked word will be interpreted as a named volume.

A relative path will cause docker to complain about the slash.

How should Scuba handle this?

1. Handle this by absolute-ifying the path.
   - Should be interpreted as relative to .scuba.yml
   - Is this necessary? Paths relative (below) .scuba.yml are already accessible via pwd.

2. Detect and complain (abort).

cc @rpasek

JonathonReinhart:
This enables the use of relative paths in a volume hostpath.

Previously, if one put this in `.scuba.yml`:
```yml
volumes:
    /foo: foo
    # or
    /foo:
        hostpath: foo
```
`foo` would be interpreted by Docker as a *named volume*, resulting in an empty directory at `/foo`, which is certainly not what the user intended.

Or, if one tried to work around that problem and do this:
```yml
volumes:
    /foo: ./foo
```
Then Docker would complain about the use of a relative path (a specifier containing a `/` but not starting with `/`), and suggest the use of an absolute path.

Now, when scuba encounters a relative path in a volume `hostpath`, it is interpreted as **relative to the scuba root** (the directory in which `.scbua.yml` is found). To avoid ambiguity with the old behavior, *relative paths must start with `./` or `../`*.

I also considered interpreting relative paths as relative to the current working directory, but that seemed like a much less useful approach:
- The common scenario would be to mount a project folder somewhere else in the filesystem:
  ```yml
  volumes:
    /var/cualqomm/license:
        hostpath: ./cq_license
  ```
- I can't think of a scenario where I would want a bind mount (to a fixed container-path) to change depending on where my current directory is within my project tree.
  - If this was truly desired, one could use `$PWD/foo`.
- The scheme whereby relative paths in config files are relative to the directory containing the config file is intuitive and widely-used:
  - In Scuba
    - `!from_yaml` ([ref](https://scuba.readthedocs.io/en/latest/configuration.html#accessing-external-yaml-content:~:text=The%20path%20of%20an%20external%20YAML%20file%20(relative%20to%20.scuba.yaml)))
  - In `.gitconfig` ([ref](https://git-scm.com/docs/git-config#:~:text=If%20the%20value%20of%20the%20variable%20is%20a%20relative%20path%2C%20the%20path%20is%20considered%20to%20be%20relative%20to%20the%20configuration%20file%20in%20which%20the%20include%20directive%20was%20found))
  - In ADMan config ([ref](https://adman.readthedocs.io/en/latest/configuration.html#types:~:text=path%20%E2%80%93%20A%20YAML%20string%3B%20the%20path%20to%20a%20file.%20The%20path%20can%20either%20be%20absolute%20(start%20with%20a%20leading%20/)%2C%20or%20relative%20to%20the%20directory%20containing%20the%20config.yml%20file.))
  - In Docker Compose:
    - `env_file` ([ref](https://docs.docker.com/compose/compose-file/05-services/#env_file:~:text=Relative%20path%20are%20resolved%20from%20the%20Compose%20file%27s%20parent%20folder.))
    - `extends.file` ([ref](https://docs.docker.com/compose/compose-file/05-services/#extends:~:text=Relative%20path.%20This%20path%20is%20considered%20as%20relative%20to%20the%20location%20of%20the%20main%20Compose%20file.))
    - service volumes ([ref](https://docs.docker.com/compose/compose-file/05-services/#short-syntax-5:~:text=the%20relative%20path%20is%20resolved%20from%20the%20Compose%20file%E2%80%99s%20parent%20directory))

Fixes #218

### DIFF
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -145,8 +145,8 @@ style <https://yaml.org/spec/1.2/spec.html#id2788097>`_:
 ``volumes``
 -----------
 
-The optional ``volumes`` node *(added in v2.9.0)* allows additional `volumes
-<https://docs.docker.com/storage/volumes/>`_ or bind-mounts to be specified.
+The optional ``volumes`` node *(added in v2.9.0)* allows additional
+`bind-mounts<https://docs.docker.com/storage/bind-mounts/>`_ to be specified.
 ``volumes`` is a mapping (dictionary) where each key is the container-path.
 In the simple form, the value is a string, the host-path to be bind-mounted:
 
@@ -154,6 +154,7 @@ In the simple form, the value is a string, the host-path to be bind-mounted:
 
     volumes:
       /var/lib/foo: /host/foo
+      /var/lib/bar: ./bar
 
 In the complex form, the value is a mapping which must contain a ``hostpath``
 subkey. It can also contain an ``options`` subkey with a comma-separated list
@@ -178,6 +179,11 @@ directory into the container at the same path.
 
 If a referenced environment variable is not set, Scuba exits with a
 configuration error.
+
+Volume container paths must be absolute.
+
+Volume host paths can be absolute or relative. If a relative path is used, it
+is interpreted as relative to the directory in which ``.scuba.yml`` is found.
 
 Volume host directories which do not already exist are created as the current
 user before creating the container.
--- a/docs/configuration.rst
+++ b/docs/configuration.rst
@@ -184,6 +184,7 @@ Volume container paths must be absolute.
 
 Volume host paths can be absolute or relative. If a relative path is used, it
 is interpreted as relative to the directory in which ``.scuba.yml`` is found.
+To avoid ambiguity, relative paths must start with ``./`` or ``../``.
 
 Volume host directories which do not already exist are created as the current
 user before creating the container.
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -332,7 +332,7 @@ def _get_volumes(
 
     vols = {}
     for cpath_str, v in voldata.items():
-        cpath = _expand_path(cpath_str)
+        cpath = _expand_path(cpath_str)  # no base_dir; container path must be absolute.
         vols[cpath] = ScubaVolume.from_dict(cpath, v, scuba_root)
     return vols
 
@@ -405,7 +405,7 @@ class ScubaVolume:
         if isinstance(node, str):
             return cls(
                 container_path=cpath,
-                host_path=_expand_path(node),
+                host_path=_expand_path(node, scuba_root),
             )
 
         # Complex form
@@ -419,7 +419,7 @@ class ScubaVolume:
                 raise ConfigError(f"Volume {cpath} must have a 'hostpath' subkey")
             return cls(
                 container_path=cpath,
-                host_path=_expand_path(hpath),
+                host_path=_expand_path(hpath, scuba_root),
                 options=_get_delimited_str_list(node, "options", ","),
             )
 
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -344,7 +344,8 @@ def _expand_path(in_str: str, base_dir: Optional[Path] = None) -> Path:
     the host environment.
 
     After environment variable expansion, absolute paths are returned as-is.
-    Relative paths are joined to base_dir, if provided.
+    Relative paths must start with ./ or ../ and are joined to base_dir, if
+    provided.
 
     Args:
       in_str: Input path as a string.
@@ -357,13 +358,14 @@ def _expand_path(in_str: str, base_dir: Optional[Path] = None) -> Path:
       ValueError: If base_dir is provided but not absolute.
       ConfigError: If a referenced environment variable is not set.
       ConfigError: An environment variable reference could not be parsed.
+      ConfigError: A relative path does not start with "./" or "../".
       ConfigError: A relative path is given when base_dir is not provided.
     """
     if base_dir is not None and not base_dir.is_absolute():
         raise ValueError(f"base_dir is not absolute: {base_dir}")
 
     try:
-        output = expand_env_vars(in_str)
+        path_str = expand_env_vars(in_str)
     except KeyError as ke:
         # pylint: disable=raise-missing-from
         raise ConfigError(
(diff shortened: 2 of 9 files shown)


JonathonReinhart:
I think I want to restrict the relative paths to starting with `./` (or `../`), just like [Docker Compose does](https://docs.docker.com/compose/compose-file/05-services/#short-syntax-5:~:text=with%20an%20error.-,To%20avoid%20ambiguities%20with%20named%20volumes%2C%20relative%20paths%20should%20always%20begin,.,-Long%20syntax).

While the long form is pretty unambiguous (given that it says `hostpath`):
```
volumes:
  /foo:
    hostpath: foo
```

The short form is currently (unintentionally, and undocumented) interpreted as a named volume.
```
volumes:
  /foo: foo  # Is this a relative path or a named volume?
```

Requiring it to start with a `./` (in both cases) makes it 100% clear it is a relative path.

I'd rather start with this strict approach and relax as necessary, rather than starting with it loosely defined and trying to tighten it up later.

xanarin:
I think the behavioral choices here make a lot of sense. I agree that a named volume is not what I would expect from having a bare directory name, and I like that `scuba` does accept a bare directory name either and instead forces the user to use `./` or `../` to formally confirm their intention. I also agree that those relative paths should be based on the `.scuba.yml` directory. I think basing in on CWD would just be a footgun.

I think in the worst case, this feature will be made less frustrating to an unsuspecting user because of #195. If a non-existent relative host path is created, it will at least be created with permissions such that the user will easily be able to delete it without having to escalate to `root`. But I think that scenario is very unlikely since this is a newly-supported option.

docs/configuration.rst
```diff
@@ -179,6 +180,10 @@ directory into the container at the same path.
 If a referenced environment variable is not set, Scuba exits with a
 configuration error.
 
+Volume host paths can be absolute or relative. If a relative path is used, it
+is interpreted as relative to the directory in which ``.scuba.yml`` is found.
+To avoid ambiguity, relative paths must start with ``./`` or ``../``.
+
```
xanarin, line 186: In [line 335 of config.py](https://github.com/JonathonReinhart/scuba/pull/227/files#diff-1bd8088b035e8e07af4ead68414dce30741e7cedbdef91de1a3d70e048ae1992R335), it specifies that container paths must be absolute. That seems like a useful note to put in this user-facing documentation as well.

scuba/config.py
```diff
@@ -323,21 +323,49 @@ def _get_delimited_str_list(data: CfgData, key: str, sep: str) -> List[str]:
...
-def _expand_path(in_str: str) -> Path:
+def _expand_path(in_str: str, base_dir: Optional[Path] = None) -> Path:
+    """Expand variables in a path string and make it absolute.
+
+    Environment variable references (e.g. $FOO and ${FOO}) are expanded using
+    the host environment.
+
+    After environment variable expansion, absolute paths are returned as-is.
+    Relative paths must start with ./ or ../ and are joined to base_dir, if
+    provided.
```
xanarin, line 348: I made up a stupid test to ensure that variable expansion happens before the relative path checking, and it passed! Up to you whether you want to add it to the test suite or not to prevent regressions:
```diff
diff --git a/tests/test_config.py b/tests/test_config.py
index 3b7fd43..ab2b4d6 100644
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1104,6 +1104,27 @@ class TestConfig:
             )
         self._invalid_config("Relative path must start with ./ or ../")
 
+    def test_volumes_hostpath_rel_in_env(
+        self, monkeypatch, in_tmp_path
+    ) -> None:
+        """volume definitions can contain environment variables, including relative path portions"""
+        monkeypatch.setenv("PREFIX", "./")
+        with open(".scuba.yml", "w") as f:
+            f.write(
+                r"""
+                image: na
+                volumes:
+                  /foo: ${PREFIX}/foo
+                """
+            )
+
+        config = load_config()
+        vols = config.volumes
+        assert vols is not None
+        assert len(vols) == 1
+
+        assert_vol(vols, "/foo", in_tmp_path / "foo")
+
     def test_volumes_contpath_rel(self, monkeypatch, in_tmp_path) -> None:
         with open(".scuba.yml", "w") as f:
             f.write(
```

tests/test_scuba.py
```diff
@@ -1,18 +1,23 @@
 import pytest
 import shlex
+from typing import Any
 from .utils import assert_vol
 
 from scuba.config import ScubaConfig, ConfigError, OverrideStr
 from scuba.scuba import ScubaContext
 
 
+def make_config(**data: Any) -> ScubaConfig:
```
xanarin, line 10: The reasoning for using this utility function instead of invoking `ScubaConfig` directly in each test is explained somewhat in the commit message, but I think a comment here would be a good addition as well. Especially because in its current form it's just a passthrough and seems to provide little value.

xanarin (approved):
Overall, this looks great. A few small questions and documentation changes are all the useful feedback I can give. I also love the additional documentation on the exceptions that a function can raise! Between that and the new type annotations the codebase is definitely getting easier to spot-check and reason about.

JonathonReinhart:
> I also love the additional documentation on the exceptions that a function can raise!

You can thank the [Google Python Style Guide](https://google.github.io/styleguide/pyguide.html#doc-function-raises) for that 😃.  (Not all of it I like though...)

> Between that and the new type annotations the codebase is definitely getting easier to spot-check and reason about.

Yes! It's so much easier to comprehend, particularly in the case of complex dict or tuple types. Also, `mypy` is surprisingly fast (likely due to its cache), so especially in cases like this where there was lots of plubing / refatoring, much of the initial edit-save-check cycle was done with static analysis only. Then `test_config.py` which is pretty fast. And only then did I run `test_main.py` which is slow because it actually runs Docker.

With this, and some of the Python 2 tech debt paid off (#226), I'm pretty happy with the codebase these days.

scuba/config.py
JonathonReinhart: Thanks!

```console
$ git apply will.diff
$ ./code_format.py --fix
$ git commit --author='Will Sloan <will@848.email>' --date='2023-09-15T03:13:05Z
[volume-relative-path d4a61bc] tests: Add test ensuring variable expansion happens before relative path check                                                                                                                                
 Author: Will Sloan <will@848.email>                                                                                                                                                                                                         
 Date: Fri Sep 15 03:13:05 2023 +0000                                                                                                                                                                                                        
 1 file changed, 21 insertions(+)
```

👉 d4a61bcb43d008d14813db605ed754f5143f378b

JonathonReinhart:
Thanks for the thorough review @xanarin!

haboustak:
Unfortunately, this MR introduced a regression in some code I was recently reviewing. We had projects written in Go that were storing their `GOMODCACHE` between invocations of `scuba` into a `go-mod-cache` named volume.

For what it's worth, I'm not in support of this usage of `scuba` volumes, but there's at least one instance of existing code that made use of it., and if it was written once, it was pasted at least 4 more times.

As I understand the usage, the intent was to mount the named volume at `/go/pkg/mod/cache` inside the container.

```Dockerfile
ENV GOMODCACHE /go/pkg/mod/cache
```

```yml
# requires scuba 2.10.0
volumes:
  ${HOME}/.cache/go-build: ${HOME}/.cache/go-build
  /go/pkg/mod/cache: go-mod-cache
```

JonathonReinhart:
> which is certainly not what the user intended.

Well that will teach me not to assume I understand my users.

> To avoid ambiguity with the old behavior, *relative paths must start with `./` or `../`.*

Whew, looks like this was a good idea.

Thanks for pointing this out. I'll open a bug to track the fix.

JonathonReinhart:
Opened #248.

PR #231, opened 2023-12-03
Rewrite scubainit in Rust

JonathonReinhart:


JonathonReinhart:
This was not intended to be "merged". I accidentally had the upstream branch of my local branch set to "origin/main". So when I did `git push`, it actually updated `main` which marked this PR as "merged". I force-pushed `main` back to where it was.

PR #239, opened 2023-12-30
Avoid lifetimes on EntFile{Reader,Writer}

JonathonReinhart:
This is a follow-up from https://github.com/JonathonReinhart/scuba/pull/232#discussion_r1437728384.

cc @jammiess

### DIFF
--- a/scubainit/src/entfiles.rs
+++ b/scubainit/src/entfiles.rs
@@ -86,21 +86,28 @@ impl<'a> EntLineParser<'a> {
 ////////////////////////////////////////////////////////////////////////////////
 // EntFileReader
 
-pub struct EntFileReader<'a, T> {
-    reader: BufReader<&'a File>,
+pub struct EntFileReader<T> {
+    reader: BufReader<File>,
     marker: PhantomData<T>, // T must be used
 }
 
-impl<T> EntFileReader<'_, T> {
-    pub fn new(file: &File) -> EntFileReader<T> {
+impl<T> EntFileReader<T> {
+    pub fn new(file: File) -> EntFileReader<T> {
         EntFileReader {
             reader: BufReader::new(file),
             marker: PhantomData,
         }
     }
+
+    /// Unwraps this `EntFileReader<T>`, returning the underlying File.
+    ///
+    /// Note that the position of the File is undefined.
+    pub fn into_inner(self) -> File {
+        self.reader.into_inner()
+    }
 }
 
-impl<T: Entry> Iterator for EntFileReader<'_, T> {
+impl<T: Entry> Iterator for EntFileReader<T> {
     type Item = Result<T, ReadEntryError>;
 
     fn next(&mut self) -> Option<Self::Item> {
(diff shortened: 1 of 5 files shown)


JonathonReinhart:
This doesn't currently work:
```
error[E0382]: use of moved value: `reader`
   --> src/main.rs:134:20
    |
115 |         let reader = groups::GroupFileReader::new(file);
    |             ------ move occurs because `reader` has type `scubainit::entfiles::EntFileReader<GroupEntry>`, which does not implement the `Copy` trait
116 |         for grp in reader {
    |                    ------ `reader` moved due to this implicit call to `.into_iter()`
...
134 |         let file = reader.into_inner();
    |                    ^^^^^^ value used here after move
    |
note: `into_iter` takes ownership of the receiver `self`, which moves `reader`
   --> /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/iter/traits/collect.rs:267:18
```

jammiess:
> This doesn't currently work:
> 
> ```
> error[E0382]: use of moved value: `reader`
>    --> src/main.rs:134:20
>     |
> 115 |         let reader = groups::GroupFileReader::new(file);
>     |             ------ move occurs because `reader` has type `scubainit::entfiles::EntFileReader<GroupEntry>`, which does not implement the `Copy` trait
> 116 |         for grp in reader {
>     |                    ------ `reader` moved due to this implicit call to `.into_iter()`
> ...
> 134 |         let file = reader.into_inner();
>     |                    ^^^^^^ value used here after move
>     |
> note: `into_iter` takes ownership of the receiver `self`, which moves `reader`
>    --> /rustc/a28077b28a02b92985b3a3faecf92813155f1ea1/library/core/src/iter/traits/collect.rs:267:18
> ```

This is pretty simple to fix. Rust is defaulting to calling `into_iter` on the reader to turn it into an iterator. However, that method consumes the iterator so that it can't be used anymore. If you make the reader mutable and then change for for loop into something like
```rs
for value in &mut reader { ... }
```
then rust won't consume the iterator while looping over it and you can continue to use the value afterwards.

JonathonReinhart:
> This is pretty simple to fix. Rust is defaulting to calling `into_iter` on the reader to turn it into an iterator. However, that method consumes the iterator so that it can't be used anymore. If you make the reader mutable and then change for for loop into something like
> 
> ```rust
> for value in &mut reader { ... }
> ```
> 
> then rust won't consume the iterator while looping over it and you can continue to use the value afterwards.

Perfect! Thanks for taking the time to explain why this didn't work and how to fix it.

> I don't know the exact rules for when rust will take ownership of iterators like this. But when I run into an issue like this I try to turn the iterator into a mutable reference and see if that fixes the issue.

Yeah, I think I'll probably have to bang my head on this sort of thing a few times before it fully sets in. But now that I understand the fix, this part of the error message makes more sense:

> ```
> `reader` moved due to this implicit call to `.into_iter()`
> ```

Sticking `&mut` there makes sense, as it avoids the move.

I'm merging this into the branch of the other PR.

PR #250, opened 2024-03-22
Add explicit support for mounting named volumes

### LINKED ISSUE #248: Restore ability to use named volumes
Prior to #227, you could, as a not-officially-support feature map a named volume at a container path:
```yaml
volumes:
  /foo: foo-volume
```

```console
$ scuba --root --image debian:8.2 /bin/bash -c "echo yes > /foo/does-it-persist"
$ scuba --root --image debian:8.2 /bin/bash -c "cat /foo/does-it-persist"
yes
$ docker volume ls
DRIVER    VOLUME NAME
local     foo-volume
```

Since #227, this no longer works.
```
scuba: Config error: Relative path must start with ./ or ../: foo-volume
```

*Reported https://github.com/JonathonReinhart/scuba/pull/227#issuecomment-2005067223 by @haboustak.*

JonathonReinhart:
Named volumes were unintentionally supported prior to #227. This change restores the prior behavior while remaining unambiguiously compatible with relative bind-mounts added in #227.

This also adds support for explicit named volumes in complex form configuration.

Fixes #248

### DIFF
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -19,6 +19,8 @@ CfgData = Dict[str, CfgNode]
 Environment = Dict[str, str]
 _T = TypeVar("_T")
 
+VOLUME_NAME_PATTERN = re.compile(r"^[a-zA-Z0-9][a-zA-Z0-9_.-]+$")
+
 
 class ConfigError(Exception):
     pass
@@ -410,8 +412,13 @@ def _absoluteify_path(in_str: str, base_dir: Optional[Path] = None) -> Path:
 @dataclasses.dataclass(frozen=True)
 class ScubaVolume:
     container_path: Path
-    host_path: Path  # TODO: Optional for anonymous volume
+    host_path: Optional[Path] = None
+    volume_name: Optional[str] = None
     options: List[str] = dataclasses.field(default_factory=list)
+
+    def __post_init__(self) -> None:
+        if sum(bool(x) for x in (self.host_path, self.volume_name)) != 1:
+            raise ValueError("Exactly one of host_path, volume_name must be set")
 
     @classmethod
     def from_dict(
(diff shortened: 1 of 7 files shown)


JonathonReinhart:
cc @haboustak

scuba/config.py
```diff
@@ -399,9 +412,14 @@ def _expand_path(in_str: str, base_dir: Optional[Path] = None) -> Path:
...
 class ScubaVolume:
     container_path: Path
-    host_path: Path  # TODO: Optional for anonymous volume
+    host_path: Optional[Path] = None
+    volume_name: Optional[str] = None
     options: List[str] = dataclasses.field(default_factory=list)
 
+    def __post_init__(self) -> None:
+        if sum(bool(x) for x in (self.host_path, self.volume_name)) != 1:
+            raise ValueError("Exactly one of host_path, volume_name must be set")
```
haboustak, line 421: I really like the clarity of this error message. However, it provides a hint to issues with `yaml` de-serialization. See below for an example with multiple `name` keys defined for a single complex `volume`. All `yaml` config would really benefit from a defined schema with constraint validation.

I'm not suggesting you need to deal with this case, but it's annoying that simple typos can fail with success.

```yml
volumes:
  /go/pkg/mod/cache:
      name: go-mod-cache
      name: go-mod-caches
```
```javascript
// docker inspect 02e871026589 | jq '.[0].Mounts | map(select(.Type=="volume"))'
[
  {
    "Type": "volume",
    "Name": "go-mod-caches",
    "Source": "/var/lib/docker/volumes/go-mod-caches/_data",
    "Destination": "/go/pkg/mod/cache",
    "Driver": "local",
    "Mode": "z",
    "RW": true,
    "Propagation": ""
  }
]
```

haboustak:
LGTM.

I tested it on a handful of existing projects that used relative paths and/or named volumes and found no issues.

I also tried to break it with combinations of variables and absolute/relative paths, but was unsuccessful.

JonathonReinhart:
> LGTM.
> 
> I tested it on a handful of existing projects that used relative paths and/or named volumes and found no issues.
> 
> I also tried to break it with combinations of variables and absolute/relative paths, but was unsuccessful.

Thanks for testing! I'll probably merge this and cut a new release tonight.

I'm going to call this a minor release (2.13.0) rather than a bugfix, because named volumes were never really supported. Also there's no path to fixing 2.12.0 without essentially just reverting it to 2.11.0.

PR #255, opened 2024-05-20
scubainit: Restore the default SIGPIPE action

### LINKED ISSUE #254: scubainit ignores SIGPIPE
The new `scubainit` included in version 2.13.0 ignores `SIGPIPE`, which is then inherited by any process that scuba runs. This behavior is included as a side-effect of the Rust entrypoint. See [rust-lang/rust #62569](https://github.com/rust-lang/rust/issues/62569)

Here's a test case that was shared with me recently and demonstrates the behavior change from `2.12.0` to `2.13.0`:
```
$ scuba --image debian:12 sh -c "yes | echo abcd"
abcd
yes: standard output: Broken pipe
```

Also, to show that `SIGPIPE` is ignored in `2.13.0`:
```
$ scuba --image debian:12 cat /proc/1/status | grep "Sig"
SigQ:	1/31475
SigPnd:	0000000000000000
SigBlk:	0000000000000000
SigIgn:	0000000000001000
SigCgt:	0000000000010002
```

I think the correct behavior for scuba is to reset `SIGPIPE` to `SIG_DFL`.

JonathonReinhart:
Rust pre-main code may change the SIGPIPE disposition to ignore:
* https://github.com/rust-lang/rust/issues/62569
* https://github.com/rust-lang/rust/issues/97889

We could use the nightly compiler flag `-Zon-broken-pipe=inherit` to disable this behavior. Instead, we take the simpler route and restore the default disposition ourselves.

Fixes #254

### DIFF
--- a/scubainit/src/main.rs
+++ b/scubainit/src/main.rs
@@ -45,6 +45,8 @@ fn run_scubainit() -> Result<()> {
     setup_logging()?;
     info!("Looking Rusty!");
 
+    restore_sigpipe_default()?;
+
     let ctx = process_envvars()?;
 
     if let Some(ref user_info) = ctx.user_info {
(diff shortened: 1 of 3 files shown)


JonathonReinhart:
@haboustak @wcampbell0x2a @jammiess

scubainit/src/main.rs
```diff
@@ -368,3 +370,20 @@ fn setup_logging() -> Result<()> {
         .verbosity(verbosity)
         .init()?)
 }
+
+// Restore the default SIGPIPE action.
+// Rust pre-main code may change the SIGPIPE disposition to ignore:
+// https://github.com/rust-lang/rust/issues/62569
+// https://github.com/rust-lang/rust/issues/97889
```
haboustak, line 377: I like the detail in this comment. I might add a note that `scubainit` doesn't rely on any specific `SIGPIPE` behavior, but that signal dispositions are inherited through `execve` and that Rust's pre-main code shouldn't influence the process executed by scuba.

From `man signal(7)`
> Signal dispositions
> ...
> A child created via fork(2) inherits a copy of its parent's signal  dispositions.   During  an
> execve(2),  the  dispositions of handled signals are reset to the default; the dispositions of
> ignored signals are left unchanged. 

The comment should help to point out that this code may be unnecessary in the future if Rust's `execve` library function is changed to restore `SIGPIPE`.

JonathonReinhart: Thanks. You're right I kind of glossed over why we care. Updated the comment (mostly verbatim from your comment).

haboustak (approved):
Tested and confirmed this resolves #254.

JonathonReinhart:
This was just released in [v2.13.1](https://github.com/JonathonReinhart/scuba/releases/tag/v2.13.1).

Issue #229, opened 2023-09-20
Scuba crashes with GitLab CI image dictionary


xanarin:
Currently, if one specifies a `.scuba.yml` like this:
```yaml
aliases:
  build: !from_yaml .gitlab-ci.yml build
```
With a GitLab CI configuration file like this:
```yaml
build:
  stage: build
  image:
    name: debian:12
  script:
    - echo "Ran successfully"
```

When one runs `scuba build`, they will see the following stack trace:
```
$ scuba build
Traceback (most recent call last):
  File "/home/wsloan/.local/bin/scuba", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/wsloan/.local/pipx/venvs/scuba/lib/python3.11/site-packages/scuba/__main__.py", line 184, in main
    rc = run_scuba(scuba_args) or 0
         ^^^^^^^^^^^^^^^^^^^^^
  File "/home/wsloan/.local/pipx/venvs/scuba/lib/python3.11/site-packages/scuba/__main__.py", line 138, in run_scuba
    dive = ScubaDive(
           ^^^^^^^^^^
  File "/home/wsloan/.local/pipx/venvs/scuba/lib/python3.11/site-packages/scuba/scuba.py", line 104, in __init__
    self.__setup_native_run()
  File "/home/wsloan/.local/pipx/venvs/scuba/lib/python3.11/site-packages/scuba/scuba.py", line 294, in __setup_native_run
    ep = get_image_entrypoint(self.context.image)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wsloan/.local/pipx/venvs/scuba/lib/python3.11/site-packages/scuba/dockerutil.py", line 131, in get_image_entrypoint
    return _get_image_config(image, "Entrypoint")
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wsloan/.local/pipx/venvs/scuba/lib/python3.11/site-packages/scuba/dockerutil.py", line 114, in _get_image_config
    info = docker_inspect_or_pull(image)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/wsloan/.local/pipx/venvs/scuba/lib/python3.11/site-packages/scuba/dockerutil.py", line 82, in docker_inspect_or_pull
    return docker_inspect(image)
[...]

Issue #200, opened 2023-02-20
GitLab CI yaml includes custom tags that cannot be resolved by from_yaml 


haboustak:
GitLab has added a custom yaml tag: [!reference](https://docs.gitlab.com/ee/ci/yaml/yaml_optimization.html#reference-tags). This tag copies content from another path within the yaml document. The example use cases are to reuse script blocks in other CI jobs. Implementation in ruby is here: [lib/gitlab/ci/config/yaml/tags/reference.rb](https://gitlab.com/gitlab-org/gitlab/-/blob/436d642725ac6675c97c7e5833d8427e8422ac78/lib/gitlab/ci/config/yaml/tags/reference.rb#L8)

I've created an [example project](https://gitlab.com/haboustak/test-ci-reference/) based on the above documentation.

It seems appropriate to create a `!from_gitlab` custom tag to load GitLab-specific yaml syntax. This would use a `SafeLoader` with support for `!reference` initially. It would also provide a place for `scuba` to support some or all of GitLab's `include` syntax.

**Sample `.gitlab-ci.yml`**
```yaml
stages:
  - build

.before:
  script:
    - echo "ABCD"

build:
  stage: build
  script:
    - !reference [.before, script]
    - echo "after ABCD completes"
```

haboustak:
Two other GitLab-flavored yaml extensions from my recent memory that would be nice to have some level of  some `scuba` support: 

1. They build compound yaml documents using the [includes](https://docs.gitlab.com/ee/ci/yaml/#include) keyword
2. They have replaced yaml anchors with the [extends](https://docs.gitlab.com/ee/ci/yaml/#extends) keyword

These extensions may be unreasonable to support within `GitlabYamlLoader`, but a warning from `scuba` would be nice. For example: 
> WARN: GitLabYamlLoader: Loading content from `.gitlab-ci.yml` with unsupported `includes` keyword.

JonathonReinhart:
I think this all sounds great. I'm all for a `!from_gitlab` (`!from_gitlab_ci_yaml` or `!from_gitlab_ci` might be a bit too long).

PR #257, opened 2024-05-27
add hotfix for gitlab extensions

FunkeCoder23:
Fix crashing issue (https://github.com/JonathonReinhart/scuba/issues/229)

Adds a `not_yet_implented` func to 'handle' `!reference` tags from gitlab

I'm currently working on a proper implementation for both of these issue, but this should work as a stop-gap until then.

### DIFF
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -155,9 +155,12 @@ class Loader(yaml.SafeLoader):
         assert isinstance(obj, OverrideMixin)
         return obj
 
+    def not_yet_implemented(self, node: yaml.nodes.Node):
+        pass
 
 Loader.add_constructor("!from_yaml", Loader.from_yaml)
 Loader.add_constructor("!override", Loader.override)
+Loader.add_constructor("!reference", Loader.not_yet_implemented)
 
 
 def find_config() -> Tuple[Path, Path, ScubaConfig]:
@@ -517,7 +520,7 @@ class ScubaAlias:
             return cls(
                 name=name,
                 script=script,
-                image=node.get("image"),
+                image=_get_str(node, "image"),
                 entrypoint=_get_entrypoint(node),
                 environment=_process_environment(
                     node.get("environment"), f"{name}.environment"
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -282,6 +282,44 @@ class TestLoadConfig(ConfigTest):
         SCUBA_YML.write_text(f"image: !from_yaml {external_yml} danger")
         self.__test_load_config_safe(external_yml)
 
+    def test_load_config_image_name(self) -> None:
+        GITLAB_YML.write_text(
+            """
+            two:
+              stage: build
+              image:
+                name: dummian:8.2
+            """
+        )
+        invalid_config(
+            config_text=f"""
+            aliases:
+              two:
+                image:  !from_yaml {GITLAB_YML} two.image
+                script: rip
+            """
+        )
+
+    def test_load_config_gitlab_extensions(self) -> None:
+        GITLAB_YML.write_text(
+            """
+            include: included.yml
+
+            image: dummian:8.2
+
+            one:
+              image: !reference [another, job, somewhere]
+            
+            two:
+              extends: one
+
+            normal:
+              image: dummian:12
+            """
+        )
+
+        load_config(config_text=f"image: !from_yaml {GITLAB_YML} normal.image")
+
 
 class TestConfigHooks(ConfigTest):
     def test_hooks_mixed(self) -> None:


scuba/config.py
```diff
@@ -517,7 +520,7 @@ def from_dict(
             return cls(
                 name=name,
                 script=script,
-                image=node.get("image"),
+                image=_get_str(node, "image"),
```
JonathonReinhart, line 523: This fix is actually independent from the `!reference` constructor addition, but only the latter is mentioned anywhere.

At a minimum, please update the commit message and/or PR description to call this out explicitly. The word "image" doesn't appear anywhere but the diff.

Bonus points if you put this (and the associated test addition) into their own commit. More bonus points if it is its own PR -- since this isn't really a GitLab "extension" to yaml, it's just a part of their schema that Scuba doesn't (yet) support.

scuba/config.py
```diff
@@ -155,9 +155,12 @@ def override(self, node: yaml.nodes.Node) -> OverrideMixin:
         assert isinstance(obj, OverrideMixin)
         return obj
 
+    def not_yet_implemented(self, node: yaml.nodes.Node):
+        pass
 
 Loader.add_constructor("!from_yaml", Loader.from_yaml)
 Loader.add_constructor("!override", Loader.override)
+Loader.add_constructor("!reference", Loader.not_yet_implemented)
```
JonathonReinhart, line 163: Can you update your PR description to reference #200 which spells out in great detail what the `!reference` tag is, and how GitLab uses it?

scuba/config.py
```diff
@@ -155,9 +155,12 @@ def override(self, node: yaml.nodes.Node) -> OverrideMixin:
         assert isinstance(obj, OverrideMixin)
         return obj
 
+    def not_yet_implemented(self, node: yaml.nodes.Node):
```
JonathonReinhart, line 158: Let's name this `!reference`, following the existing pattern. That it is not yet implemented is mostly inconsequential.

scuba/config.py
```diff
@@ -155,9 +155,12 @@ def override(self, node: yaml.nodes.Node) -> OverrideMixin:
...
+        pass
```
JonathonReinhart, line 159: This implementation appears to just *ignore* the `!reference` tags, without raising an error or warning the user, simply returning `None` (implicitly). This seems like a *huge* problem to me.

Consider this example:
```yaml
.setup:
  script:
    - do-something-really-important

build:
  stage: build
  script:
    - !reference [.setup, script]
    - depends-on-important-stuff
```

The result (AFICT) is a script of `[None, "depends-on-important-stuff"]`.

I suspect that will crash somewhere later because of the `None` value.

If we pretend for a second that the None value is ignored, then the resulting script will be missing the first step, and executing anyway, despite the YAML author's intentions.

I don't see how this can be better than the current state of affairs (in which it just crashes).

scuba/config.py
JonathonReinhart, line 163: `!reference` is only added for compatibility with content imported from `.gitlab-ci.yml` (via `!from_yaml`), correct? It is not a part of the documented `.scuba.yml` schema.

So I don't think we should be adding this constructor to this `Loader` class. It should be part of a separate Loader class used only when parsing `.gitlab-ci.yml` files. And how do we know we're parsing a `.gitlab-ci.yml` file (or a differently-named file using its schema)? We should use a new tag + constructor which is a variant of `!from_yaml`.

This is all spelled out in #200.

tests/test_config.py
```diff
@@ -282,6 +282,44 @@ def test_load_config_safe_external(self) -> None:
...
+        invalid_config(
+            config_text=f"""
+            aliases:
+              two:
+                image:  !from_yaml {GITLAB_YML} two.image
+                script: rip
+            """
+        )
+
+    def test_load_config_gitlab_extensions(self) -> None:
```
JonathonReinhart, line 303: What are GitLab "extensions"? AFAICT this isn't a term that's used anywhere in YAML, Scuba, or GitLab documentation. If this is testing the `!reference` tag, this should probably have "reference" in the name.

JonathonReinhart, line 303: Many of the other tests have a docstring that briefly describes what is being verified. Please consider adding that here as well.

tests/test_config.py
```diff
@@ -282,6 +282,44 @@ def test_load_config_safe_external(self) -> None:
         SCUBA_YML.write_text(f"image: !from_yaml {external_yml} danger")
         self.__test_load_config_safe(external_yml)
 
+    def test_load_config_image_name(self) -> None:
+        GITLAB_YML.write_text(
+            """
+            two:
+              stage: build
+              image:
+                name: dummian:8.2
```
JonathonReinhart, line 291: It is not obvious what exactly is being tested here. I had to read it several times to figure it out.

A comment like this would go a long way in aiding understanding.

```python
# .gitlab-ci.yml supports a long-form `image` specification
# with `name` and other sub-keys:
# https://docs.gitlab.com/ee/ci/yaml/#image
# 
# Scuba doesn't currently support that, so make sure we handle
# this error gracefully.
#
# TODO(#230): Support this complex `image` format.
```

tests/test_config.py
```diff
@@ -282,6 +282,44 @@ def test_load_config_safe_external(self) -> None:
...
+            
+            two:
+              extends: one
+
+            normal:
+              image: dummian:12
+            """
+        )
+
+        load_config(config_text=f"image: !from_yaml {GITLAB_YML} normal.image")
```
JonathonReinhart, line 321: Most other tests go on to validate the contents of the object returned from `load_config`. If you do that, I think you'll find that the results are unsatisfactory.

JonathonReinhart (changes requested):
Thanks for the PR.

I think this has some fundamental problems that would need to be addressed before something like this could be merged.

FunkeCoder23:
I think I was trying to do to much with this one branch (not a single-issue branch). I'm going to close this in favor of a PR that fixes #229 only, and take the comments you've left as guidance for the more in-depth/proper handling of a gitlab yaml loader

PR #265, opened 2025-06-08
ci: Update actions/upload-artifact and actions/download-artifact to v4

### LINKED ISSUE #264: Migrate to upload-artifact v4
https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/

JonathonReinhart:
Not affected by the breaking changes.

Fixes #264

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -56,7 +56,7 @@ jobs:
 
     - name: Upload Unit Test Results
       if: always()
-      uses: actions/upload-artifact@v3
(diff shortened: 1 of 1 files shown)


JonathonReinhart:
There are multiple failures
* Python 3.7 is no longer supported (#242)
* mypy is upset about the `pytest.raises()` return value (#266)

To keep this PR focused, I will land it with those failures while I dig out of this three-foot hole.

The fact that any tests passed indicates that the migration in this PR is correct.

PR #271, opened 2025-06-21
Use type hinting generics from standard collections

JonathonReinhart:
This eliminates all uses of these members of `typing`:  
`Tuple`, `List`, `Dict`, `Set`, `FrozenSet`, `Type`


JonathonReinhart:
This was new in Python 3.9 (#243)

PR #274, opened 2025-07-25
Parse SCUBAINIT_UMASK as an octal-encoded string

### LINKED ISSUE #273: scubainit may be setting an incorrect umask
After a recent upgrade to `scuba 1.14.0` I began noticing that my build products are missing the "other users" read permission bit. As a result, files created inside the container were only readable by the current user and root.

```
$ scuba --version
scuba 2.14.0

$ docker --version
Docker version 28.1.1, build 4eba377

$ head -n1 /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 12 (bookworm)"

$ scuba --image debian:12 touch scuba.perm.test && stat -c %a scuba.perm.test ; rm -f scuba.perm.test
640
$ scuba --image debian:12 umask
0026

$ docker run --rm -it debian:12 sh -c "touch scuba.perm.test && stat -c %a scuba.perm.test"
644
$ docker run --rm -it debian:12 sh -c "umask"
0022
```

I expected that this regression was introduced in the latest build, because I would have noticed it earlier, but it looks like it was new in `2.13.0`.
```
$ pipx install -f scuba==2.13.1
$ scuba --version
scuba 2.13.1
$ scuba --image debian:12 touch scuba.perm.test && stat -c %a scuba.perm.test ; rm -f scuba.perm.test
640
$ scuba --image debian:12 umask
0026

$ pipx install -f scuba==2.13.0
$ scuba --version
scuba 2.13.0
$ scuba --image debian:12 touch scuba.perm.test && stat -c %a scuba.perm.test ; rm -f scuba.perm.test
640
$ scuba --image debian:12 umask
0026

$ pipx install -f scuba==2.12.0
$ scuba --version
scuba 2.12.0
$ scuba --image debian:12 touch scuba.perm.test && stat -c %a scuba.perm.test ; rm -f scuba.perm.test
644
$ scuba --image debian:12 umask
0022
```

haboustak:
Fixes #273

### DIFF
--- a/scubainit/src/util.rs
+++ b/scubainit/src/util.rs
@@ -71,7 +71,7 @@ pub fn pop_env_uint(name: &str) -> Result<Option<u32>> {
     };
     let value: u32 = value_str
         .parse()
-        .context(format!("Parsing decimal integer variable {name}=\"{value_str}\""))?;
+        .with_context(|| format!("Parsing decimal integer variable {name}=\"{value_str}\""))?;
     Ok(Some(value))
 }
 
@@ -86,7 +86,7 @@ pub fn pop_env_octal(name: &str) -> Result<Option<u32>> {
         Some(s) => s,
     };
     let value = u32::from_str_radix(&value_str, 8)
-        .context(format!("Parsing octal integer variable {name}=\"{value_str}\""))?;
+        .with_context(|| format!("Parsing octal integer variable {name}=\"{value_str}\""))?;
     Ok(Some(value))
 }
 pub fn make_executable(path: &str) -> std::io::Result<()> {
--- a/scubainit/src/util.rs
+++ b/scubainit/src/util.rs
@@ -59,7 +59,7 @@ pub fn pop_env_bool(name: &str) -> bool {
     result
 }
 
-/// Gets and unsets a uint environment variable `name`, or Ok(None) if it is not set.
+/// Gets and unsets a decimal uint environment variable `name`, or Ok(None) if it is not set.
 ///
 /// # Errors
 ///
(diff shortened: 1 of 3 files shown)


scubainit/src/util.rs
```diff
@@ -75,6 +75,20 @@ pub fn pop_env_uint(name: &str) -> Result<Option<u32>> {
...
+/// # Errors
+///
+/// This function will return an error if the variable is set, but is not a valid integer string.
+pub fn pop_env_octal(name: &str) -> Result<Option<u32>> {
+    let value_str = match pop_env_str(name) {
+        None => return Ok(None),
+        Some(s) => s,
+    };
+    let value = u32::from_str_radix(&value_str, 8)
+        .context(format!("Parsing integer variable {name}=\"{value_str}\""))?;
```
james-a-johnson, line 89:
```suggestion
        .with_context(|| format!("Parsing integer variable {name}=\"{value_str}\""))?;
```

This way the string will only be allocated when the error actually occurs instead of being eagerly evaluated since it's directly a parameter to the method.

scubainit/src/util.rs
```diff
@@ -71,10 +71,24 @@ pub fn pop_env_uint(name: &str) -> Result<Option<u32>> {
...
+/// Gets and unsets an octal uint environment variable `name`, or Ok(None) if it is not set.
+///
+/// # Errors
+///
+/// This function will return an error if the variable is set, but is not a valid integer string.
+pub fn pop_env_octal(name: &str) -> Result<Option<u32>> {
+    let value_str = match pop_env_str(name) {
+        None => return Ok(None),
+        Some(s) => s,
+    };
```
wcampbell0x2a, line 87: > Reading Rust is becoming as much fun for me as reading Perl!

Time to embrace pattern matching ;) You could use [let-else](https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html) here.

```suggestion
    let Some(value_str) = pop_env_str(name) else {
        return Ok(None);
```

The previous `pop_env_uint` could use the same pattern.

JonathonReinhart: Ah, I do love `let-else` these days, but I was very new to Rust when I first wrote this file.

We can leave this as-is for consistency for and I may do an update later.

JonathonReinhart:
Thanks @haboustak for tracking this down and implementing the fix!

PR #277, opened 2026-01-24
Add `!from_yaml` multi-document support

### LINKED ISSUE #276: `!from_yaml` doesn't support multi-document YAML files
Take the following YAML and Scuba config as an example:

```yaml
# sample.yml
spec:
  inputs:
    user:
      default: 'test-user'
---
job:
  image: debian
  script:
    - echo "An example script section."
```

```yaml
# .scuba.yml
image: !from_yaml sample.yml job.image
aliases:
  test: cat /etc/os-release
```

Then, `scuba test` gives:
```$ scuba test
scuba: Config error: Error loading .scuba.yml: expected a single document in the stream
  in "sample.yml", line 1, column 1
but found another document
  in "sample.yml", line 5, column 1
```

GitLab requires user-defined pipeline input variables to be provided [in the first document of the CI file](https://docs.gitlab.com/ci/yaml/#header-keywords:~:text=The%20header%20must%20be%20at%20the%20top%20of%20the%20file%2C%20separated%20from%20the%20rest%20of%20the%20configuration%20with%20%2D%2D%2D%2E) (the `---` causes two separate documents to be created); however, [Scuba's usage of `yaml.load()`](https://github.com/JonathonReinhart/scuba/blob/1c65adc872aaf1443cbf7f91eb95f0f659868189/scuba/config.py#L115) assumes a single document.

brxnnan:
Fixes #276

### DIFF
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -1,17 +1,17 @@
 from __future__ import annotations
+
 import dataclasses
 import os
-from pathlib import Path
 import re
 import shlex
-from typing import Any, Optional, TextIO, TypeVar, Union
-from typing import overload
+from pathlib import Path
+from typing import Any, Optional, TextIO, TypeVar, Union, overload
 
 import yaml
 import yaml.nodes
 
-from .constants import DEFAULT_SHELL, SCUBA_YML
 from . import utils
+from .constants import DEFAULT_SHELL, SCUBA_YML
 from .dockerutil import make_vol_opt
 
 CfgNode = Any
--- a/scuba/config.py
+++ b/scuba/config.py
@@ -59,7 +59,7 @@ class OverrideStr(str, OverrideMixin):
 # http://stackoverflow.com/a/9577670
 class Loader(yaml.SafeLoader):
     _root: Path  # directory containing the loaded document
-    _cache: dict[Path, Any]  # document path => document
+    _cache: dict[Path, list[Any]]  # doc path => list of YAML documents from that file
 
     def __init__(self, stream: TextIO):
         if not hasattr(self, "_root"):
@@ -109,20 +109,39 @@ class Loader(yaml.SafeLoader):
         path = self._root / filename
 
         # Load the other YAML document
-        doc = self._cache.get(path)
-        if not doc:
+        docs = self._cache.get(path)
+        if not docs:
             with path.open("r") as f:
-                doc = yaml.load(f, self.__class__)
-                self._cache[path] = doc
+                docs = list(yaml.load_all(f, self.__class__))
+                self._cache[path] = docs
 
         # Retrieve the key
         try:
-            cur = doc
             # Use a negative look-behind to split the key on non-escaped '.' characters
-            for k in re.split(r"(?<!\\)\.", key):
-                cur = cur[
-                    k.replace("\\.", ".")
-                ]  # Be sure to replace any escaped '.' characters with *just* the '.'
+            # Be sure to replace any escaped '.' characters with *just* the '.'
+            top_key, *keys = [
+                key.replace("\\.", ".") for key in re.split(r"(?<!\\)\.", key)
+            ]
+
+            # Ensure the top key is unique between all documents
+            found_top = [
+                (i, doc[top_key])
+                for i, doc in enumerate(docs, start=1)
+                if isinstance(doc, dict) and top_key in doc
+            ]
+            if not found_top:
+                raise KeyError
+            elif len(found_top) > 1:
+                matching_docs = ", ".join([str(found[0]) for found in found_top])
+                raise yaml.YAMLError(
+                    f"Key {key!r} found multiple times in {filename} in documents {matching_docs}"
+                )
+            else:
+                cur = found_top[0][1]
+
+            # Now find it
+            for k in keys:
+                cur = cur[k]
         except KeyError:
             raise yaml.YAMLError(f"Key {key!r} not found in {filename}")
         return cur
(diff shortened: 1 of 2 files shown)


scuba/config.py
```diff
@@ -59,7 +59,7 @@ class OverrideStr(str, OverrideMixin):
 # http://stackoverflow.com/a/9577670
 class Loader(yaml.SafeLoader):
     _root: Path  # directory containing the loaded document
-    _cache: dict[Path, Any]  # document path => document
+    _cache: dict[Path, list[Any]]  # document path => document
```
JonathonReinhart: I think this comment needs to be updated, e.g.:
```
    _cache: dict[Path, list[Any]] # file path => list of YAML documents from that file
```

scuba/config.py
```diff
@@ -109,20 +109,41 @@ def from_yaml(self, node: yaml.nodes.Node) -> Any:
...
-                    k.replace("\\.", ".")
-                ]  # Be sure to replace any escaped '.' characters with *just* the '.'
+            # Be sure to replace any escaped '.' characters with *just* the '.'
+            top_key, *keys = [
+                key.replace("\\.", ".") for key in re.split(r"(?<!\\)\.", key)
+            ]
+
+            # Ensure the top key is unique between all documents
+            found_top = [
+                (i + 1, doc[top_key]) for i, doc in enumerate(docs) if top_key in doc
```
JonathonReinhart: Instead of `i + 1` you can pass `start=1` to `enumerate()` which I think is a little clearer.

scuba/config.py
```diff
@@ -109,20 +109,41 @@ def from_yaml(self, node: yaml.nodes.Node) -> Any:
...
-        if not doc:
+        docs = self._cache.get(path)
+        if not docs:
             with path.open("r") as f:
-                doc = yaml.load(f, self.__class__)
-                self._cache[path] = doc
+                docs = [
+                    doc
+                    for doc in yaml.load_all(f, self.__class__)
+                    if isinstance(doc, dict)
```
JonathonReinhart: By skipping non-dict documents here, you could end up emitting a confusing error message below, because the document indices won't match.

E.g.

```yaml
# Document 1: A Mapping (Dictionary)
foo: "bar"
---
# Document 2: A Scalar (Not a mapping)
"This second document is just a simple string, not a dictionary."
---
# Document 3: A Mapping (Dictionary)
foo: "bar"
```

attempting to look up `foo` would result in an error message like
```
Key 'foo' found multiple times in example.yml in documents 1, 2"
```
when it was actually in documents 1, 3.

The right solution might be to skip the filtering here, and do it below.

Or maybe insert a dummy empty dict here, e.g.
```python
                docs = [
                    # Ignore non-mapping documents.
                    doc if isinstance(doc, dict) else {}
                    for doc in yaml.load_all(f, self.__class__)
                ]
```

JonathonReinhart (changes requested):
This looks good, thanks!

I pointed out a couple of minor things that should be easy to resolve.

scuba/config.py
brxnnan: Fixed with https://github.com/JonathonReinhart/scuba/pull/277/commits/7e3df91db46ee7a8c1caf68f01eecd43e9eb7dd1

scuba/config.py
brxnnan: Nice, I didn't know `enumerate` had that. Fixed with https://github.com/JonathonReinhart/scuba/pull/277/commits/7e3df91db46ee7a8c1caf68f01eecd43e9eb7dd1

scuba/config.py
brxnnan: Good catch. I think it makes more sense to have the cache's values be the full list of documents (the type would have to be updated to be `dict[Path, list[dict]]` instead if replaced with `{}` anyways). Updated to filter it later with https://github.com/JonathonReinhart/scuba/pull/277/commits/7e3df91db46ee7a8c1caf68f01eecd43e9eb7dd1

scuba/config.py
```diff
@@ -329,13 +348,11 @@ def _get_typed_val(
 
 
 @overload  # When default is None, can return None (Optional).
-def _get_str(data: CfgData, key: str, default: None = None) -> Optional[str]:
-    ...
+def _get_str(data: CfgData, key: str, default: None = None) -> Optional[str]: ...
```
JonathonReinhart: It looks like your editor is formatting differently than this project's standard.

Can you please [run code formatting](https://github.com/JonathonReinhart/scuba/blob/main/CONTRIBUTING.md#code-format)?

JonathonReinhart (changes requested):

scuba/config.py
brxnnan: Oops. My editor was using a newer version of `black` which was outside of my venv. Fixed with https://github.com/JonathonReinhart/scuba/pull/277/commits/72f1cea71cca8384c86a18fc4bb0f3d98ec4b33d and https://github.com/JonathonReinhart/scuba/pull/277/commits/c6d18b44fa2199f934ac82eb7a72189418a4f381

JonathonReinhart:
Thanks @brxnnan for the contribution!

How urgently are you looking for this to be included in a new release?

brxnnan:
> How urgently are you looking for this to be included in a new release?

It'd be put to use immediately, so if it isn't inconvenient for you, sooner rather than later would be appreciated.

JonathonReinhart:
You got it.

[v2.14.2](https://github.com/JonathonReinhart/scuba/releases/tag/v2.14.2)

brxnnan:
Thank you!

scuba/config.py
JonathonReinhart: FYI I just updated `black` to 2026 in #282.

PR #286, opened 2026-06-17
Add aarch64 support to scubainit

retu2libc:
Implement multi-architecture support for scubainit.

- build scubainit for the aarch64 musl target
- scuba detects which scubainit to use based on image architecture since some hosts (MacOS) can run containers for architectures other than what their cpu supports.
- fix cross-target linking by configuring rust-lld for musl targets
- add rustup target setup and multi-arch build steps to CONTRIBUTING.mg

This was mostly put together so that I can locally use scuba for development on my aarch64 machine. Would be really cool to get builds for aarch64 up on pypi though.

Previously:
```sh
$ scuba --image alpine:latest
exec /.scuba/scubainit: exec format error
```
Now:
```sh
~$ scuba --image alpine:latest
~ $ uname -a
Linux 26650320bf60 6.12.90+deb13.1-arm64 #1 SMP Debian 6.12.90-2 (2026-05-27) aarch64 Linux
```

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -8,10 +8,11 @@ on:
 jobs:
   build-and-test:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
-        python-version: ["3.10", "3.11", "3.12", "3.13"]
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
+        python-version: ["3.10", "3.11", "3.12", "3.13", "3.14"]
       fail-fast: False
 
     steps:
@@ -58,7 +59,7 @@ jobs:
       if: always()
       uses: actions/upload-artifact@v4
       with:
-        name: Unit Test Results (Python ${{ matrix.python-version }})
+        name: Unit Test Results (${{ matrix.runner }}, Python ${{ matrix.python-version }})
         path: unit-test-results.xml
 
   # https://github.com/marketplace/actions/publish-unit-test-results#use-with-matrix-strategy
--- a/.github/workflows/rust-check.yml
+++ b/.github/workflows/rust-check.yml
@@ -7,7 +7,10 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
 
     steps:
     - name: Checkout code
--- a/Makefile
+++ b/Makefile
@@ -1,15 +1,19 @@
 .PHONY: all
-all: scuba/scubainit
+all: scuba/scubainit-all
+
+SCUBAINIT_BIN_TARGETS := x86_64-unknown-linux-musl aarch64-unknown-linux-musl
 
 .PHONY: scubainit  # Defer dependency-tracking to Cargo
 scubainit:
 	make -C $@ test scubainit
 
-# Copy the binary into the scuba python package
-scuba/scubainit: scubainit
-	cp $</scubainit $@
+# Copy all architecture variants into the scuba python package
+scuba/scubainit-all: scubainit
+	for target in $(SCUBAINIT_BIN_TARGETS); do \
+		cp scubainit/scubainit-$$target scuba/scubainit-$$target; \
+	done
 
 .PHONY: clean
 clean:
 	make -C scubainit clean
-	rm -f scuba/scubainit
+	rm -f scuba/scubainit-*
--- a/scubainit/.cargo/config.toml
+++ b/scubainit/.cargo/config.toml
@@ -1,0 +1,5 @@
+[target.x86_64-unknown-linux-musl]
+linker = "rust-lld"
+
+[target.aarch64-unknown-linux-musl]
+linker = "rust-lld"
--- a/setup.py
+++ b/setup.py
@@ -66,7 +66,7 @@ setup(
     packages=["scuba"],
     package_data={
         "scuba": [
-            "scubainit",
+            "scubainit-*",
         ],
     },
     include_package_data=True,  # https://github.com/pypa/setuptools/issues/1064
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,10 +6,12 @@ information you desire.*
 
 1. Download and install [docker-ce](https://docs.docker.com/engine/install/debian/#install-using-the-repository)
 2. Download and install [rust](https://www.rust-lang.org/tools/install)
+	1. Run `rustup default stable`
+	2. Run `rustup target add x86_64-unknown-linux-musl aarch64-unknown-linux-musl`
 <!-- Specific python version? -->
 3. Download and install python3 and pip
 4. Run `source ./dev_bootstrap.sh`
-5. Run `test-docker-images/build_all.sh`
+5. Run `./test-docker-images/build_all.sh`
 
 ## Build and test
 
--- a/scuba/.gitignore
+++ b/scuba/.gitignore
@@ -1,1 +1,1 @@
-scubainit
+scubainit-*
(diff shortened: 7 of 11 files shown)


.github/workflows/build-test.yml
```diff
@@ -8,9 +8,10 @@ on:
 jobs:
   build-and-test:
 
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
     strategy:
       matrix:
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
```
JonathonReinhart, line 14: If we're going to pin the `arm` build to 24.04, we should pin the host (x86-64) build to 24.04 also.

.github/workflows/publish-main.yml
```diff
@@ -34,11 +41,13 @@ jobs:
...
         CI_VERSION_BUILD_NUMBER: ${{ github.run_id }}
-      run: python -m build
+      run: python -m build --wheel
 
     - name: Publish
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_TEST_USERNAME }}
         TWINE_PASSWORD: ${{ secrets.PYPI_TEST_PASSWORD }}
-      run: twine upload --repository-url https://test.pypi.org/legacy/ dist/*
+      run: twine upload --repository-url https://test.pypi.org/legacy/ dist/*.whl
```
JonathonReinhart: Why are we changing this?

.github/workflows/publish-release.yml
```diff
@@ -30,10 +37,12 @@ jobs:
...
+        SCUBAINIT_TARGET: ${{ matrix.scubainit-target }}
+        SCUBA_WHEEL_PLAT_NAME: ${{ matrix.wheel-plat-name }}
+      run: python -m build --wheel
 
     - name: Publish
       env:
         TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
         TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-      run: twine upload dist/*
+      run: twine upload dist/*.whl
```
JonathonReinhart: Same question here as on `publish-main.yml`.

.github/workflows/rust-check.yml
```diff
@@ -7,7 +7,10 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        runner: [ubuntu-latest, ubuntu-24.04-arm]
```
JonathonReinhart, line 13: Same comment here as on `.github/workflows/build-test.yml`.

scubainit/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[target.x86_64-unknown-linux-musl]
+linker = "rust-lld"
```
JonathonReinhart, line 2: What does this do differently than before?

The commit message just says 

> - fix cross-target linking by configuring rust-lld for musl targets

but it's not clear what was broken or how this fixes it.

Makefile
```diff
@@ -1,13 +1,24 @@
 .PHONY: all
 all: scuba/scubainit
 
+HOST_ARCH := $(shell uname -m)
+ifeq ($(HOST_ARCH),aarch64)
+DEFAULT_SCUBAINIT_TARGET := aarch64-unknown-linux-musl
+else ifeq ($(HOST_ARCH),arm64)
+DEFAULT_SCUBAINIT_TARGET := aarch64-unknown-linux-musl
+else
+DEFAULT_SCUBAINIT_TARGET := x86_64-unknown-linux-musl
```
JonathonReinhart: I think it would be better to add a condition for `x86_64` rather than default to it.

And then the `else` should be a fatal error "Unrecognized machine: $(HOST_ARCH)"

JonathonReinhart: Of course, we don't really care about `DEFAULT_SCUBAINIT_TARGET` at all if `SCUBAINIT_TARGET` is set so it shouldn't be fatal in that case

Makefile
```diff
@@ -1,13 +1,24 @@
...
+SCUBAINIT_TARGET ?= $(DEFAULT_SCUBAINIT_TARGET)
+
 .PHONY: scubainit  # Defer dependency-tracking to Cargo
 scubainit:
 	make -C $@ test scubainit
 
 # Copy the binary into the scuba python package
 scuba/scubainit: scubainit
-	cp $</scubainit $@
+	cp scubainit/scubainit-$(SCUBAINIT_TARGET) $@
```
JonathonReinhart: This appears to copy only one file. I thought this added dynamic multi-arch support, not just support for aarch64 hosts?

.github/workflows/build-test.yml
JonathonReinhart, line 1: (GitHub doesn't allow comment on the commit message, so I'm adding it here.)

* Scuba doesn't use this `feat(x)` syntax. Please follow the format in the commit history.
* I think it would be wise to separate the `scubainit` build changes from the CI changes (if they are logically separable).
* The commit author is set to `user <user>` -- please configure your git client.

setup.py
```diff
@@ -36,6 +37,15 @@ def run(self):
         super().run()
 
 
+class bdist_wheel(bdist_wheel):
+    def finalize_options(self):
+        super().finalize_options()
+        self.root_is_pure = False
+        plat_name = os.getenv("SCUBA_WHEEL_PLAT_NAME")
+        if plat_name:
+            self.plat_name = plat_name
```
JonathonReinhart: Scuba doesn't currently set `plat_name`.

As a result, its published [wheel files](https://pypi.org/project/scuba/2.14.2/#files) have their platform tag set to `any` and are named e.g. `scuba-2.14.2-py3-none-any.whl`.

One could argue this is currently incorrect, since it contains just one x86_64 `scubainit` and -- as you noted in your PR description -- trying to use scuba on a Mac results in `exec format error`.

However, this PR claims to support true multi-arch support (runtime detection) -- which would presumably mean that the scuba wheel should continue to advertise itself as `any` platform, since it performs the platform detection at runtime.

So the choice to set `plat_name` here seems very inconsistent with that goal.

setup.py
```diff
@@ -36,6 +37,15 @@ def run(self):
         super().run()
 
 
+class bdist_wheel(bdist_wheel):
+    def finalize_options(self):
+        super().finalize_options()
+        self.root_is_pure = False
+        plat_name = os.getenv("SCUBA_WHEEL_PLAT_NAME")
```
JonathonReinhart: It makes little sense, to me, to set `plat_name` based only on an external environment variable when one could just use the `--plat-name` command line option instead, which does the same thing.

Please take a look at [how staticx does this](https://github.com/JonathonReinhart/staticx/blob/main/setup.py). But also look at the comment below and determine if this is even necessary or desirable.

.github/workflows/publish-main.yml
JonathonReinhart, line 1: (This is a comment on the commit message)

I wouldn't call this a "refactor" when it changes behavior.

.github/workflows/publish-main.yml
```diff
@@ -11,13 +11,6 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
-    strategy:
```
JonathonReinhart: Okay this commit undoes or redoes the majority of the changes from the previous commit, rendering most of review irrelevant. I will have to review this another time.

JonathonReinhart (changes requested):
After carefully reviewing the first commit, I find that the second commit undoes or redoes a bunch of those changes rendering my review wasted.

Please update or resubmit a more coherent PR and I will re-review.

retu2libc:
i had to change it because if scabs i nested the host system binary per
original commit it would not have worked correctly. the second commit is
largely ai. as a result.

JonathonReinhart:
> i had to change it because if scabs i nested the host system binary per original commit it would not have worked correctly. the second commit is largely ai. as a result.

If the approach changes, then please amend the commits accordingly, rather than piling on contradictory changes.

Every commit in a PR should be one relevant piece of a coherent story evolving the codebase. PR commits shouldn't be a record of your local development story. This is why `git rebase -i` exists.

